PR #2, opened 2021-05-04
Move incubator/hnc to root

rjbez17:
This PR is one of two. It does nothing more than:

1) Delete all root files and directories from multi-tenancy
2) Copy the original (generated from template) repo files back into the root (like owners files)
3) Move incubator/hnc into the root directory


yiqigao217 (commented):
/lgtm

PR #31, opened 2021-05-06
Fix e2e tests

adrianludwin:
The e2e tests refer to the old "incubator/hnc" directory which doesn't
exist in this repo.

Tested: as usual, I don't have any way to test Prow configs or flows.
However, this is in response to a Prow failure that clearly identifies
this line.


yiqigao217 (commented):
/lgtm

Issue #9, opened 2021-05-05
Replace excluded-namespace label with included-namespace


yiqigao217:
<a href="https://github.com/yiqigao217"><img src="https://avatars.githubusercontent.com/u/51385444?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [yiqigao217](https://github.com/yiqigao217)**
_Friday Apr 30, 2021 at 16:22 GMT_
_Originally opened as https://github.com/kubernetes-sigs/multi-tenancy/issues/1503_

----

Consider excluding namespaces with K8s 1.21+ namespace metadata.name label - 
See https://github.com/kubernetes/kubernetes/issues/92157#issuecomment-825160958

yiqigao217:
<a href="https://github.com/yiqigao217"><img src="https://avatars.githubusercontent.com/u/51385444?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [yiqigao217](https://github.com/yiqigao217)**
_Friday Apr 30, 2021 at 20:07 GMT_

----

Another thought is that instead of using `excluded`, maybe use `included`? Considering the damage it could cause if the system namespaces are not labeled as excluded.

HNC can add the `included` label to all namespaces except those flagged as excluded-namespaces in HNC `container.args`. Then our object webhook can remain fail CLOSE and use that label instead.

yiqigao217:
(The `app.kubernetes.io/managed-by` label may be a good option, but ACM already uses it on namespaces created from the repo. In that case, when HNC works with ACM in the same cluster, HNC actually "manages" the namespace but cannot have the `app.kubernetes.io/managed-by: hnc.x-k8s.io` on that namespace.)

Since this label is mainly for the object webhook to decide whether to enforce policy on the namespaces, we may name it as **`hnc.x-k8s.io/webhooks-enforced`**?

The `excluded-namespace` container args in the HNC deployment manifest is the only source-of-truth. HNC won't reconcile these namespaces or objects in these namespaces.

As for the currently used `hnc.x-k8s.io/excluded-namespace` label introduced in v0.8, we will depreciate it in v0.9 since this is a non-breaking change.

WDYT @rjbez17 @adrianludwin

rjbez17:
Sorry, I never hit submit last week. Yes, I do agree with this approach. Would we use a mutating WH on the NS object to add the label? Do it in reconciliation later? I assume it will also disallow removing the label/re-add it?

[...]

PR #36, opened 2021-05-10
Replace excluded namespace label with included

yiqigao217:
Part of #9
Next PR would be updating the label concept doc.

Missing `excluded-namespace` label on system namespaces would break the
entire cluster. Instead of always trying hard to make sure the
`excluded-namespace` label are set, this PR replaces it with included
namespace concept (`included-namespace` label added by HC reconciler on
all namespaces except those listed as excluded in the HNC container
args).

Remove the `excluded-namespace` label. Add the `included-namespace` label.
Update HC reonciler to always set the `included-namespace` label to true
on non-excluded namespaces and remove it on excluded namespaces. Update
namespace webhook to block adding/updating the new label improperly.

Update the Makefile to remove the `excluded-namespace` label setting.

Tested by `make test` and `make test-e2e`.

### DIFF
--- a/internal/reconcilers/hierarchy_config_test.go
+++ b/internal/reconcilers/hierarchy_config_test.go
@@ -381,26 +381,31 @@ var _ = Describe("Hierarchy", func() {
 		Eventually(getLabel(ctx, nms[5], nms[0]+api.LabelTreeDepthSuffix)).Should(Equal("3"))
 	})
 
-	It("should remove excluded namespace labels from non-excluded namespaces", func() {
+	It("should remove included-namespace namespace labels from excluded namespaces", func() {
 		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
-		l := map[string]string{api.LabelExcludedNamespace: "true"}
-
-		// Set excluded namespace labels on foo and bar. We are not verifying the
-		// labels are added here because our reconciler will remove it and the
-		// verification would be flaky.
-		fooName := createNSWithLabel(ctx, "foo", l)
-		barName := createNSWithLabel(ctx, "bar", l)
-
 		kubeSystem := getNamespace(ctx, "kube-system")
+
+		// Add additional label "other:other" to verify the labels are updated.
+		l := map[string]string{api.LabelIncludedNamespace: "true", "other": "other"}
 		kubeSystem.SetLabels(l)
 		updateNamespace(ctx, kubeSystem)
-		Eventually(getLabel(ctx, "kube-system", api.LabelExcludedNamespace)).Should(Equal("true"))
+		// Verify the labels are updated on the namespace.
+		Eventually(getLabel(ctx, "kube-system", "other")).Should(Equal("other"))
+		// Verify the included-namespace label is removed by the HC reconciler.
+		Eventually(getLabel(ctx, "kube-system", api.LabelIncludedNamespace)).Should(Equal(""))
+	})
 
-		// Verify the excluded namespace label is removed from both foo and bar, but
-		// is not removed from kube-system.
-		Eventually(getLabel(ctx, fooName, api.LabelExcludedNamespace)).Should(Equal(""))
-		Eventually(getLabel(ctx, barName, api.LabelExcludedNamespace)).Should(Equal(""))
-		Eventually(getLabel(ctx, "kube-system", api.LabelExcludedNamespace)).Should(Equal("true"))
+	It("should set included-namespace namespace labels properly on non-excluded namespaces", func() {
+		// Create a namespace without any labels.
+		fooName := createNS(ctx, "foo")
+		// Verify the label is eventually added by the HC reconciler.
+		Eventually(getLabel(ctx, fooName, api.LabelIncludedNamespace)).Should(Equal("true"))
+
+		l := map[string]string{api.LabelIncludedNamespace: "false"}
+		// Create a namespace with the label with a wrong value.
+		barName := createNSWithLabel(ctx, "bar", l)
+		// Verify the label is eventually updated to have the right value.
+		Eventually(getLabel(ctx, barName, api.LabelIncludedNamespace)).Should(Equal("true"))
 	})
 })
 
--- a/internal/validators/namespace.go
+++ b/internal/validators/namespace.go
@@ -76,7 +76,7 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 
 	switch req.op {
 	case k8sadm.Create:
-		if rsp := v.illegalExcludedNamespaceLabel(req); !rsp.Allowed {
+		if rsp := v.illegalIncludedNamespaceLabel(req); !rsp.Allowed {
 			return rsp
 		}
 		// This check only applies to the Create operation since namespace name
@@ -85,7 +85,7 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 			return rsp
 		}
 	case k8sadm.Update:
-		if rsp := v.illegalExcludedNamespaceLabel(req); !rsp.Allowed {
+		if rsp := v.illegalIncludedNamespaceLabel(req); !rsp.Allowed {
 			return rsp
 		}
 		// This check only applies to the Update operation. Creating a namespace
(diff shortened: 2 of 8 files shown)


internal/validators/namespace.go
```diff
@@ -106,17 +106,34 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
-			// even when the VWHConfiguration is removed, adding the label to a non-
-			// excluded namespace would pass but the label is immediately removed; when
-			// the VWHConfiguration is there but the reconcilers are down, any request
-			// gets denied anyway.
-			msg := fmt.Sprintf("You cannot exclude this namespace using the %q label. See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#excluded-namespace-label for detail.", api.LabelExcludedNamespace)
+// Todo add label concept and add `See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#included-namespace-label for detail`
+//  to the message below. Github issue - https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/9
+func (v *Namespace) illegalIncludedNamespaceLabel(req *nsRequest) admission.Response {
+	val, found := req.ns.Labels[api.LabelIncludedNamespace]
+	excludedNS := config.ExcludedNamespaces[req.ns.Name]
```
adrianludwin, line 113: Nit: `isExcluded` reads more easily. `excludedNS` makes it look like it should contain the string value of an excluded namespace.

internal/validators/namespace.go
```diff
@@ -106,17 +106,34 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
-			// label should have already been removed by our reconciler. For example,
-			// even when the VWHConfiguration is removed, adding the label to a non-
-			// excluded namespace would pass but the label is immediately removed; when
-			// the VWHConfiguration is there but the reconcilers are down, any request
-			// gets denied anyway.
-			msg := fmt.Sprintf("You cannot exclude this namespace using the %q label. See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#excluded-namespace-label for detail.", api.LabelExcludedNamespace)
+// Todo add label concept and add `See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#included-namespace-label for detail`
+//  to the message below. Github issue - https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/9
+func (v *Namespace) illegalIncludedNamespaceLabel(req *nsRequest) admission.Response {
+	val, found := req.ns.Labels[api.LabelIncludedNamespace]
```
adrianludwin, line 112: Nit: `val` and `found` are a bit vague for later in this function. Maybe `hasLabel` and `labelValue`?

internal/reconcilers/hierarchy_config_test.go
```diff
@@ -381,26 +381,30 @@ var _ = Describe("Hierarchy", func() {
...
-		fooName := createNSWithLabel(ctx, "foo", l)
-		barName := createNSWithLabel(ctx, "bar", l)
-
 		kubeSystem := getNamespace(ctx, "kube-system")
+
+		l := map[string]string{api.LabelIncludedNamespace: "true"}
 		kubeSystem.SetLabels(l)
 		updateNamespace(ctx, kubeSystem)
-		Eventually(getLabel(ctx, "kube-system", api.LabelExcludedNamespace)).Should(Equal("true"))
+		// Verify the label is added after updating the namespace.
```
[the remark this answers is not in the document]
yiqigao217: Yes. I added an addition label to verify the label updates.

internal/validators/namespace.go
```diff
@@ -106,17 +106,34 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+	excludedNS := config.ExcludedNamespaces[req.ns.Name]
+	switch req.op {
+	case k8sadm.Create:
+		if found {
+			msg := fmt.Sprintf("You cannot add the %q label. Only HNC can add this label to namespaces.", api.LabelIncludedNamespace)
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
```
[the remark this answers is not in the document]
yiqigao217: Agreed. Updated the logic and all the test cases.

internal/validators/namespace.go
```diff
@@ -106,20 +106,49 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
-			// label should have already been removed by our reconciler. For example,
-			// even when the VWHConfiguration is removed, adding the label to a non-
-			// excluded namespace would pass but the label is immediately removed; when
-			// the VWHConfiguration is there but the reconcilers are down, any request
-			// gets denied anyway.
-			msg := fmt.Sprintf("You cannot exclude this namespace using the %q label. See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#excluded-namespace-label for detail.", api.LabelExcludedNamespace)
+// illegalIncludedNamespaceLabel checks if there's any illegal use of the
+// included-namespace label on namespaces.
+func (v *Namespace) illegalIncludedNamespaceLabel(req *nsRequest) admission.Response {
+	labelValue, hasLabel := req.ns.Labels[api.LabelIncludedNamespace]
```
adrianludwin, line 112: Maybe file this as a TODO and a future issue, but I think I'd also like to see this validator allow any request that doesn't _change_ these values, even if they're wrong. E.g. if this is an included namespace but the label is already missing, don't reject the request just because it continues to be missing. Just let the reconciler fix it up.

If you do want to fix this up, you'll have to add `nsOld` to `nsRequest` and then you'll say something like:

```
// Early exit if nothing has changed
if req.nsOld != nil {
  oldVal, oldHas := rq.oldNS.Labels[api.LabelIncludedNamespace]
  if oldVal == labelValue && oldHas == hasLabel {
    return allow("")
  }
}
```

internal/validators/namespace_test.go
```diff
@@ -222,38 +228,49 @@ func setSubAnnotation(ns *corev1.Namespace, pnm string) {
...
-		{name: "deny creating illegal", nsInst: illegalInst, op: k8sadm.Create, fail: true},
-		{name: "allow deleting illegal", nsInst: illegalInst, op: k8sadm.Delete},
+		{name: "allow creating non-excluded namespace without label", op: k8sadm.Create},
+		{name: "allow creating non-excluded namespace with label (e.g. from a yaml file)", lval: "true", op: k8sadm.Create},
+		{name: "deny creating non-excluded namespace with label with wrong value(e.g. from a yaml file)", lval: "else", op: k8sadm.Create, fail: true},
+		{name: "allow creating excluded namespace without label", excludedNS: true, op: k8sadm.Create},
+		{name: "deny creating excluded namespace with label", excludedNS: true, lval: "true", op: k8sadm.Create, fail: true},
+
+		{name: "allow updating non-excluded namespace with correct label (which is actually set by HC reconciler)", lval: "true", op: k8sadm.Update},
+		{name: "deny updating non-excluded namespace without label", op: k8sadm.Update, fail: true},
```
adrianludwin, line 250: Does this mean that if I say "kubectl apply -f ns.yaml" twice on the same file, it could pass the first time (if the ns doesn't exist yet) and fail the second time? That doesn't really seem right. Lots of people and systems (include Config Sync) automatically say "apply -f" and expect it to work, and if the config didn't know about our labels, this will break them.

Maybe we should just allow it and trust the reconciler to fix things up? It does open up a very brief security window but perhaps we should plug that with a mutating admission webhook, not a validating one. Maybe file an issue to turn this into a mutating webhook?

yiqigao217: Agreed. Updated to allow and filed #37 .

internal/validators/namespace.go
adrianludwin: That's interesting. I agree with you that this is will happen in practice if the reconciler is working properly. However, that's a pretty strong coupling - to understand why _this_ code works correctly, you need to fully understand how a _different_ piece of code works. If that other code changes (because of a bug or because of an intentional change), this code could break. Plus it's just harder to understand.

On the other hand, if you early-exit and `return true` if these values are unchanged, now this code works _in isolation_. You don't have to care at all about what the reconciler's doing - the reconciler is now basically a fully redundant backup to this code, and as long as either of them are working as intended, HNC will continue to work.

Does that make sense?

internal/validators/namespace.go
```diff
@@ -106,20 +106,44 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+	// are down, any request gets denied anyway.
+	if isExcluded && hasLabel {
+		// Todo add label concept and add `See https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#included-namespace-label for detail`
+		//  to the message below. Github issue - https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/9
+		msg := fmt.Sprintf("You cannot enforce webhook rules on this excluded namespace using the %q label.", api.LabelIncludedNamespace)
+		return deny(metav1.StatusReasonForbidden, msg)
 	}
+
+	// An included-namespace should not have the included-namespace label with a
+	// value other than "true" at any time. We will allow updating an included
```
adrianludwin, line 132: Mention the issue you've filed here

internal/validators/namespace.go
yiqigao217: I agree on the decoupling of code in the reconciler and the validator here. But maybe it's better to implement it in the MWH and drop these VWH rules entirely?

adrianludwin: That sounds reasonable too at this point if it's all going to be temporary.

yiqigao217: I'm kinda hesitant on whether we should use VWH or MWH.

Please CMIIW - seems if we compare the ns labels in the VWH, we have to call the API server once to get the namespace object. MWH saves us 1 api call since it updates the ns object in the request directly. If it's the case, seem MWH is the better option.

WDYT @rjbez17  ?
(anyway if we decide to replace it with MWH, I will make the changes in the next PR.)

adrianludwin: I'm good with that, thanks.

adrianludwin (commented):
/lgtm
/hold
/assign @rjbez17

yiqigao217:
/hold cancel

@rjbez17 Please feel free to take a look. FYI I will make the VWH rules on namespaces `included-namespace` label in this PR into MWH in the next PR (see discussion https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/36#discussion_r632806124)

PR #39, opened 2021-05-18
Add MWH and update VWH on included-namespace label

### LINKED ISSUE #37: Consider using a mutating webhook on included namespaces
Part of #9
See https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/36#discussion_r631420215

Since people or system may apply namespace yaml file with no `included-namespace` labels on the namespaces, the attempts after the first time will be rejected if our webhook doesn't allow updating included namespaces without `included-namespace` label. Thus in #36, the webhook allows updating included namespace without the `included-namespace` label.

The HC reconciler would add the `included-namespace` label immediately, but there's a small gap in between. So we can consider using a mutating wehbook for that.

yiqigao217:
Part of #9.
Fixes #37.

Add mutating webhook to set the correct `included-namespace` label on
non-excluded namespaces if it's missing. Leave all other situation as is
and let the validating webhook to do the validation.

Update namespace validator to allow all other changes if the illegal
included-namespace label is unchanged. Since namespace mutator ensures
the label exists on included namespaces, update rules to ignore the case
when the label is missing on the included namespaces.

Add old instance field to the namespace validator. Decode it from
request to compare with the new instance. Add more test cases since we
now have more combination of cases to test.

Update the namespace tree label update logs and function name in the HC
reconciler to make it clear that the HC reconciler only updates the tree
labels and won't update included-namespace label if the MWH works fine.

Tested by `make test` with new test cases. Also tested manually by
applying namespace manifests with different labels, viewed the logs and
verified the MWH and VWH work fine.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -1,6 +1,34 @@
 
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-namespace
+  failurePolicy: Ignore
+  name: namespacelabel.hnc.x-k8s.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - namespaces
+  sideEffects: None
+
+---
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   creationTimestamp: null
@@ -8,7 +36,6 @@ metadata:
 webhooks:
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: webhook-service
@@ -29,7 +56,6 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: webhook-service
@@ -50,7 +76,6 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: webhook-service
@@ -72,7 +97,6 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: webhook-service
@@ -94,7 +118,6 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: webhook-service
--- a/internal/reconcilers/hierarchy_config.go
+++ b/internal/reconcilers/hierarchy_config.go
@@ -298,7 +298,7 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	r.syncConditions(log, inst, ns, deletingCRD, hadCrit)
 
 	// Sync the tree labels. This should go last since it can depend on the conditions.
-	nsCustomerLabelUpdated := r.syncLabel(log, nsInst, ns)
+	nsCustomerLabelUpdated := r.syncTreeLabels(log, nsInst, ns)
 
 	return initial || nsCustomerLabelUpdated
 }
@@ -461,8 +461,8 @@ func (r *HierarchyConfigReconciler) syncAnchors(log logr.Logger, ns *forest.Name
 	}
 }
 
-// Sync namespace tree labels and other labels. Return true if the labels are updated.
-func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) bool {
+// Sync namespace tree labels. Return true if the labels are updated.
+func (r *HierarchyConfigReconciler) syncTreeLabels(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) bool {
 	if ns.IsExternal() {
 		metadata.SetLabel(nsInst, nsInst.Name+api.LabelTreeDepthSuffix, "0")
 		return false
@@ -504,7 +504,7 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 	// Update the labels in the forest so that we can quickly access the labels and
 	// compare if they match the given selector
 	if ns.SetLabels(nsInst.Labels) {
-		log.Info("Namespace labels have been updated.")
+		log.Info("Namespace tree labels have been updated.")
 		return true
 	}
 	return false
--- a/internal/validators/namespace.go
+++ b/internal/validators/namespace.go
@@ -24,7 +24,7 @@ const (
 // Note: the validating webhook FAILS CLOSE. This means that if the webhook goes down, all further
 // changes are forbidden.
 //
-// +kubebuilder:webhook:admissionReviewVersions=v1;v1beta1,path=/validate-v1-namespace,mutating=false,failurePolicy=fail,groups="",resources=namespaces,sideEffects=None,verbs=delete;create;update,versions=v1,name=namespaces.hnc.x-k8s.io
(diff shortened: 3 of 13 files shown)


internal/mutator/namespace_included_label.go
```diff
@@ -0,0 +1,77 @@
...
+// be more easily unit tested (ie without constructing a full admission.Request).
+func (n *NamespaceIncludedLabel) handle(log logr.Logger, ns *corev1.Namespace) {
+	isExcluded := config.ExcludedNamespaces[ns.Name]
+
+	// Ensure the labels map is not nil and remove the included-namespace label if
+	// it exists.
+	if ns.Labels == nil {
+		ns.Labels = map[string]string{}
+	}
+	delete(ns.Labels, api.LabelIncludedNamespace)
```
adrianludwin, line 62: I think it might be worth distinguishing between labels that are _missing_ or _wrong_. If a label is missing, we should add it, but if a label is _wrong_, we should reject it rather than silently change it. E.g. if `kube-system` has `included-namespace`, I think we should just reject that; if another namespace has `included-namespace=cat`, we should reject that as well. What you've written here is certainly _correct_ but I think it would be surprising, since people expect defaults to be set, but not for their explicitly created values to change.

You can either keep that logic in the validating webhook if you like (that's kind of thematically nice) and _only_ add in the missing labels here, or you could reject bad labels here as well. Up to you, but the first option will be a smaller diff.

Note that the validating webhooks always run _after_ the mutating webhooks so you can make the validating webhooks very strict (e.g. the label must exist and must have the correct value).

internal/reconcilers/hierarchy_config.go
```diff
@@ -461,8 +461,8 @@ func (r *HierarchyConfigReconciler) syncAnchors(log logr.Logger, ns *forest.Name
 	}
 }
 
-// Sync namespace tree labels and other labels. Return true if the labels are updated.
```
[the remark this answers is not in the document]
yiqigao217: Right. I just updated the log message to say `tree` labels specifically.

config/webhook/manifests.yaml
```diff
@@ -1,4 +1,33 @@
...
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  - v1beta1
```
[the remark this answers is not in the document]
yiqigao217: Since we [explicitly](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#install-or-upgrade-hnc-on-a-cluster) say HNC requires K8s 1.16 or higher, where v1 is introduced, it's okay to remove v1beta1. I'm going to remove all v1beta1 from `admissionReviewVersions`.

Back to your question - I don't know if we still support `v1beta1`. Is it only possible for us to support both `v1` and `v1beta1` if we have a conversion webhook? Does `admissionReviewVersions: v1, v1beta1` mean if v1 is not supported by the apiserver, we need a conversion webhook to convert v1 to v1beta1 for the old apiserver?

From https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/, it says 
> Webhooks are required to support at least one AdmissionReview version understood by the current and previous API server.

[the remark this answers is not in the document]
adrianludwin: This is v1beta1 of the admission review objects, not of the HNC API. It could be that controller-runtime actually handles both v1beta1 and v1, and puts whatever it gets into the `admission.Request` object that gets passed to validator. I'm not sure though.

Maybe it's fine to leave this as-is for now and file a new issue to clean this up later.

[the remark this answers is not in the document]
adrianludwin: Looks like controller-runtime does accept both v1beta1 and v1, since they're identical: https://github.com/kubernetes-sigs/controller-runtime/pull/1284

internal/mutator/namespace_included_label.go
yiqigao217: SG. I will only mutate included namespace with missing label to have the label and restore the VWH rules.

BTW in that case, we still need a client in the namespace VWH to get the namespace instance to compare the instance vs the request to deny the request if the labels are **newly**-added on excluded namespace or **newly**-modified on the included namespace, right?

adrianludwin: Discussed offline - the Request will include the old copy of the namespace inst.

config/webhook/manifests.yaml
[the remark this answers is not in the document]
yiqigao217: I still removed v1beta1 for the reason mentioned above:
> Since we explicitly say HNC requires K8s 1.16 or higher, where v1 is introduced, it's okay to remove v1beta1. I'm going to remove all v1beta1 from admissionReviewVersions.

internal/validators/namespace.go
```diff
@@ -34,8 +34,10 @@ type Namespace struct {
 
 // nsRequest defines the aspects of the admission.Request that we care about.
 type nsRequest struct {
-	ns *corev1.Namespace
-	op k8sadm.Operation
+	nnm   string
+	newns *corev1.Namespace
+	oldns *corev1.Namespace
```
adrianludwin, line 39: Rather than splitting the existing `ns` into three variables, I'd recommend adding `oldNS` and _only_ populating it for updates. E.g. even for delete operations, you put the instance in `ns`, not `oldNS`. That preserves a lot more existing code and would make this change smaller and easier to validate (I talk a bit more about this in DecodeRequest, below).

internal/validators/namespace.go
```diff
@@ -107,60 +109,71 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
-	// existing illegal included-namespace label should have already been removed
-	// by our reconciler. For example, even when the VWHConfiguration is removed,
-	// adding the label to an excluded namespace would pass but the label is
-	// immediately removed; when the VWHConfiguration is there but the reconcilers
-	// are down, any request gets denied anyway.
+	labelValue, hasLabel := req.newns.Labels[api.LabelIncludedNamespace]
+	var oldLabelValue string
+	var oldHasLabel bool
+	if req.oldns != nil {
+		oldLabelValue, oldHasLabel = req.oldns.Labels[api.LabelIncludedNamespace]
```
adrianludwin, line 119: I would have probably just put this here:

```
if req.oldns != nil && req.oldns.Labels[api.LabelIncludedNamespace] == labelValue {
  return allow("")
}
```

Then, after this line, you know that something's changed so you could leave the rest of the logic alone. Up to you if you want to revert to that.

internal/validators/namespace.go
```diff
@@ -203,28 +218,41 @@ func (v *Namespace) illegalCascadingDeletion(ns *forest.Namespace) admission.Res
 // decodeRequest gets the information we care about into a simple struct that's easy to both a) use
 // and b) factor out in unit tests.
 func (v *Namespace) decodeRequest(log logr.Logger, in admission.Request) (*nsRequest, error) {
```
adrianludwin, line 220: I think the changes (and probably the overall complexity) of this function could be reduced if you keep the existing structure. E.g.:

* if op is delete, decode `in.OldObject.Raw` into `ns`
* else, decode the request into `ns`, and decode `in.OldObject.Raw` into `oldNS` if it's non-nil

We don't support K8s 1.14 anymore so you can take out the special-case code for that.

This preserves all existing behaviour (and code) and _only_ modifies the code when oldNS is relevant. Even most of your changes to the existing tests would go away.

adrianludwin (commented):
lgtm but with one

config/webhook/manifests.yaml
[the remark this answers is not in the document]
adrianludwin: sg

internal/validators/namespace.go
yiqigao217: Hmm, why we allow when `req.oldns.Labels[api.LabelIncludedNamespace] == labelValue`? The requested (new) namespace instance can change to have no/wrong label for included namespace or have label on excluded namespace, which should both be rejected.

yiqigao217: Discussed offline, I mistook the `labelValue` as the correct value.
Adrian's suggestion sgtm, will make that update.

yiqigao217:
Please take another look. Meanwhile, I will make the manual test into e2e tests in the next PR.

adrianludwin (commented):
/lgtm
/approve
/cc @rjbez17 

Hey Ryan, today's Yiqi's last day working on this so I'm just going to approve it. If you have any concerns I can make the changes next week. Thanks!

PR #45, opened 2021-05-20
Multi arch image builds

RealHarshThakur:
Fixes https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/21

Notes to reviewer/maintainer:
* Secret named GCR_KEY needs to be created for this repo which should contain the Service Account token   
* Moving forward, we need to have branches with release names(v0.9, etc) and the image tag will reflect that

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,0 +1,81 @@
+name: release
+
+on: 
+  create:
+  push:
+    branches:
+      - master
+      - 'v*'
+    tags:
+      - 'v*'
+
+# TODO: Migrate building binaries from Cloud Build
+jobs:
+  hnc-manager:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Set Release Tag
+        run: |
+          TAG="${GITHUB_REF#refs/*/v}"
+          echo "RELEASE_TAG=${TAG}" >> $GITHUB_ENV
+
+      - name: Set Build Date
+        id: date
+        run: |
+          echo "::set-output name=HNC_BUILD_TIMESTAMP::$(date -u +'%Y-%m-%dT%H:%M:%S%Z')"
+
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+        # GCR_KEY is the service account key in json which is base64 encoded and stored in the secrets section of the repo.
+          service_account_key: ${{ secrets.GCR_KEY }}
+          export_default_credentials: true
+        
+      # Configure docker to use the gcloud command-line tool as a credential helper
+      - run: |
+          # Set up docker to authenticate
+          # via gcloud command-line tool.
+          gcloud auth configure-docker
+
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v3
+        with:
+          images: gcr.io/k8s-staging-multitenancy/hnc-manager
+          tags: |
+            type=raw,value=multiarch-{{branch}}
+  
+      - name: Print Tags
+        run: |
+          echo "${{ steps.docker_meta.outputs.tags }}"
+          echo "RELEASE TAG: ${RELEASE_TAG}"
+
+    ## QEMU is an emulator which makes it possible to transalte instruction sets to different platforms.
+    ## In this step, we setup the driver which Docker can use to build images on different platforms
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: v0.5.1
+
+
+      - name: Build & Push Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          platforms: linux/amd64, linux/arm64, linux/arm/v7
+          tags: |
+            ${{ steps.docker_meta.outputs.tags }}
+          build-args: |
+            DBUILD_DATE=${{ steps.date.outputs.HNC_BUILD_TIMESTAMP }}
+            DBUILD_REPO_URL=https://github.com/kubernetes-sigs/hierarchical-namespaces
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -9,7 +9,7 @@ on:
     tags:
       - 'v*'
 
-# TODO: Migrate building binaries from Cloud Build
+# TODO: Migrate building manifests and kubectl plugins from Cloud Build
 jobs:
   hnc-manager:
     runs-on: ubuntu-latest
@@ -25,12 +25,13 @@ jobs:
       - name: Set Build Date
         id: date
         run: |
-          echo "::set-output name=HNC_BUILD_TIMESTAMP::$(date -u +'%Y-%m-%dT%H:%M:%S%Z')"
+          echo "::set-output name=HNC_BUILD_TIMESTAMP::$(date -u +'%Y-%m-%dT%H-%M-%S')"
 
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@master
         with:
         # GCR_KEY is the service account key in json which is base64 encoded and stored in the secrets section of the repo.
+        # We think that the Service account only requires the permission to push to the required registry but we're not certain
           service_account_key: ${{ secrets.GCR_KEY }}
           export_default_credentials: true
         
@@ -46,7 +47,7 @@ jobs:
         with:
           images: gcr.io/k8s-staging-multitenancy/hnc-manager
           tags: |
-            type=raw,value=multiarch-{{branch}}
+            type=raw,value=multiarch-{{branch}}-${{ steps.date.outputs.HNC_BUILD_TIMESTAMP }}
   
       - name: Print Tags
         run: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY cmd/ cmd/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
+RUN CGO_ENABLED=0 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
 
 # Copied from kubebuilder scaffold to run as nonroot at
 # https://github.com/kubernetes-sigs/kubebuilder/blob/7af89cb00c224c57ece37dc14ea37caf1eb769db/pkg/scaffold/v2/dockerfile.go#L60
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,14 @@ build: generate fmt vet manifests
 	     -o bin/kubectl/kubectl-hns_darwin_amd64 \
 	     -ldflags="-X sigs.k8s.io/hierarchical-namespaces/internal/version.Version=${HNC_IMG_TAG}" \
 	     ./cmd/kubectl/main.go
+	GOOS=linux GOARCH=arm64 go build \
+	     -o bin/kubectl/kubectl-hns_linux_arm64 \
+	     -ldflags="-X sigs.k8s.io/hierarchical-namespaces/internal/version.Version=${HNC_IMG_TAG}" \
+	     ./cmd/kubectl/main.go
+	GOOS=linux GOARCH=arm go build \
+	     -o bin/kubectl/kubectl-hns_linux_arm \
+	     -ldflags="-X sigs.k8s.io/hierarchical-namespaces/internal/version.Version=${HNC_IMG_TAG}" \
+	     ./cmd/kubectl/main.go
 
 # Clean all binaries (manager and kubectl)
 clean: krew-uninstall
@@ -231,6 +239,22 @@ endif
 docker-build: generate fmt vet
 	@echo "Warning: this does not run tests. Run 'make test' to ensure tests are passing."
 	docker build . -t ${HNC_IMG}
+
+
+buildx-setup:
+	## This script needs to be run to start the emulator for multiarch builds
+	# This driver translates the instruction set to different platforms 
+	docker run --rm --privileged linuxkit/binfmt:v0.8
+	docker buildx create --use --name=qemu
+	docker buildx inspect --bootstrap
+
+
+# Build and push multi-arch image
+docker-push-multi: buildx-setup generate fmt vet
+	@echo "Warning: this does not run tests. Run 'make test' to ensure tests are passing."
+	docker buildx build \
+	--push \
+	--platform linux/arm64,linux/amd64,linux/arm/v7  --tag ${HNC_IMG} .
 
 ###################### KIND ACTIONS #########################
 
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.4.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
         args:
         # The value of --upstream must match the value of --metrics-addr passed
         # to the manager binary in the base config in


adrianludwin:
Sorry for the long delay, reviewing this now

adrianludwin:
> Secret named GCR_KEY needs to be created for this repo which should contain the Service Account token

@RealHarshThakur what does "secret" mean in this context? Does Github have its own concept of secrets? Sorry, I don't know the GH flow.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
+name: release
+
+on:
+  push:
+    branches:
+      - master
```
adrianludwin, line 6: Where do the artifacts go when this is invoked on the master branch?

I wouldn't mind adding the `v*` branches as well (not just the tags)

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+    tags:
+      - v*
+
+jobs:
+  hnc-manager:
```
adrianludwin, line 11: I don't see anything here to build the manifests or plugins. Is the goal to add those later? They don't need to be in the first PR.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Set Release Tag
+        run: |
+          TAG="${GITHUB_REF#refs/*/v}"
```
adrianludwin, line 19: What does this give you if you're running on `master`?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+          echo "RELEASE_TAG=${TAG}" >> $GITHUB_ENV
+
+      - name: Set Build Date
+        id: date
+        run: |
+          echo "::set-output name=DATE::$(date -u +'%Y-%m-%dT%H:%M:%S%Z')"
```
adrianludwin, line 25: Where is this echo'd to? When does `::set-output` actually get run?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+    # Setup gcloud CLI
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
adrianludwin, line 28: This step doesn't appear to have a `name` or `id`; some have one but not the other. Is this intentional?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+      # Configure docker to use the gcloud command-line tool as a credential helper
+      - run: |
+          # Set up docker to authenticate
+          # via gcloud command-line tool.
+          gcloud auth configure-docker
+
+      - name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1
```
adrianludwin, line 42: What does this do? Is there a more official (e.g. Github-sanctioned or Google-sanctioned) action to do the same thing?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+          # add each registry to which the image needs to be pushed here
+          images: |
+            gcr.io/k8s-staging-multitenancy/hnc-manager
+          tag-latest: true
+          tag-semver: |
+            {{version}}
+
+      - name: Print Tags
+        run: |
+          echo "${{ steps.docker_meta.outputs.tags }}"
```
adrianludwin, line 53: Where does "tags" come from? I see `tag-latest` and `tag-semver`; does `crazy-max/ghaction-docker-meta` create an output called `tags`?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,79 @@
...
+          echo "RELEASE TAG: ${RELEASE_TAG}"
+
+      - name: Set up QEMU
```
adrianludwin, line 56: Can you comment why we need qemu?

Makefile
```diff
@@ -232,6 +240,15 @@ docker-build: generate fmt vet
 	@echo "Warning: this does not run tests. Run 'make test' to ensure tests are passing."
 	docker build . -t ${HNC_IMG}
 
+
+# Build and push multi-arch image
+multiarch: generate fmt vet
```
adrianludwin, line 245: Maybe use the existing pattern, e.g. `docker-push-multi` and not `multiarch`?

RealHarshThakur:
Yes, GitHub can store secrets and only the owner of the repo can set it.  https://docs.github.com/en/actions/reference/encrypted-secrets

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.4.0
+        image: quay.io/brancz/kube-rbac-proxy:v0.10.0
```
adrianludwin, line 21: I'm not sure about this one. I trust the kubebuilder build process; I don't know who `quay.io/brancz` is or if we should trust their images. I'd prefer something like `make manifests-multi` so that people have to explicitly opt into using this version of kube-rbac-proxy from an untrusted source. Or even just tell people to modify the manifest themselves (e.g. change `gcr.io/k8s-staging-multitenancy/hnc-manager:v0.9.0` to `gcr.io/k8s-staging-multitenancy/hnc-manager:v0.9.0-multi`, and change `gcr.io/kubebuilder/kube-rbac-proxy:v0.4.0` to `quay.io/brancz/kube-rbac-proxy:v0.10.0` _at your own risk_).

Is there a bug against kubebuilder to produce multiarch images?

hack/buildx.sh
```diff
@@ -0,0 +1,5 @@
+## This script needs to be run to start the emulator for multiarch builds
```
adrianludwin, line 1: Why not inline this into the makefile?

cloudbuild.yaml
```diff
@@ -87,8 +87,4 @@ steps:
   - '-u'
   - '$_HNC_USER:$_HNC_PERSONAL_ACCESS_TOKEN'
   - 'https://uploads.github.com/repos/$_HNC_REPO_OWNER/hierarchical-namespaces/releases/$_HNC_RELEASE_ID/assets?name=krew-kubectl-hns.yaml'
-# Build Docker image
-- name: gcr.io/cloud-builders/docker
-  args: ['build', '-t', 'gcr.io/$PROJECT_ID/$_HNC_IMG_NAME:$_HNC_IMG_TAG', 'hierarchical-namespaces']
 
-images: ['gcr.io/$PROJECT_ID/$_HNC_IMG_NAME:$_HNC_IMG_TAG']
```
adrianludwin, line 94: I'd rather keep this in for now, and have the GH Actions flow create images that are labelled differently. I don't like the idea of having two completely different flows where _both_ are necessary. I'd rather get the GH Action flow to the point where we can deprecate the Cloud Build one completely, but until then, I'd like Cloud Build to continue to be "complete."

RealHarshThakur:
Thanks for the review @adrianludwin , will work on it tomorrow/day after tomorrow.

adrianludwin:
Thanks! And I'll read up a bit more on GH actions.

.github/workflows/image.yaml
RealHarshThakur: Yes, will migrate the rest as separate PR.

config/default/manager_auth_proxy_patch.yaml
RealHarshThakur: I think the ownership has been passed and KubeBuilder maintainers are contributing to the same repo.  Here's the [repo ](https://github.com/brancz/kube-rbac-proxy) and here's a comment by one of the maintainers suggesting to use the image(https://github.com/kubernetes-sigs/kubebuilder/issues/1637#issuecomment-746035030) .

.github/workflows/image.yaml
RealHarshThakur: Yes, every step can has an input and output. We can reference the output values by referencing it's name.  For example: https://github.com/docker/metadata-action#outputs

.github/workflows/image.yaml
RealHarshThakur: It gives it as `refs/heads/master`  whereas when it's pushed to `v1`, it gives it as `1`. 

https://github.com/RealHarshThakur/hierarchical-namespaces/runs/2753272153?check_suite_focus=true#step:8:11
https://github.com/RealHarshThakur/hierarchical-namespaces/runs/2753141639?check_suite_focus=true

.github/workflows/image.yaml
RealHarshThakur: I added the branches, removed the tags. Is that alright?

.github/workflows/image.yaml
RealHarshThakur: It gets used in the last step . https://github.com/RealHarshThakur/hierarchical-namespaces/runs/2753272153?check_suite_focus=true#step:11:106

RealHarshThakur:
@adrianludwin I made the changes you recommended at most places. Replied on the others, PTAL

config/default/manager_auth_proxy_patch.yaml
adrianludwin: Ah, ok, I see that `github.com/brancz` is the source-of-truth for kube-rbac-proxy. I'm just not sure about their supply chain. Let me ask and find out why kubebuilder thought it was necessary to have their own copies in the first place.

adrianludwin: So I asked around, and apparently `gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0` is multi-architecture. I think I'd prefer to use that instead if it works.

The reason it's in `gcr.io/kubebuilder` is that this registry is owned by a Kubernetes SIG, while `quay.io` is not. This isn't an absolutely massive deal but it's why I'd prefer to follow kubebuilder's lead and use that image if it's reasonable to.

.github/workflows/image.yaml
RealHarshThakur: We could also not push image for master branch. Only when a release is made and a branch is created, we create an image for it. We can have a separate image tag for Github Actions, just hope it doesn't confuse the end users.  WDYT?

.github/workflows/image.yaml
RealHarshThakur: Added tags too.

RealHarshThakur:
@adrianludwin sorry for the delay. I think I addressed most comments, PTAL.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,81 @@
...
+on: 
+  create:
+  push:
+    branches:
+      - master
+      - 'v*'
+    tags:
+      - 'v*'
+
+# TODO: Migrate building binaries from Cloud Build
```
adrianludwin, line 12: s/binaries/manifests and kubectl plugins/

.github/workflows/image.yaml
adrianludwin: This applies to the release branches as well. It's _only_ Git tags that only get built once.

I'd prefer if we could append some unique prefix onto each build that comes from a branch (and not a tag). Failing that, can we say `-latest` on it so people know not to rely on it being constant?

Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY cmd/ cmd/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
+RUN CGO_ENABLED=0 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
```
adrianludwin, line 16: Do we need to add these two env vars somewhere else? To the makefile, perhaps?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Set Build Date
+        id: date
+        run: |
+          echo "::set-output name=HNC_BUILD_TIMESTAMP::$(date -u +'%Y-%m-%dT%H:%M:%S%Z')"
+
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+        # GCR_KEY is the service account key in json which is base64 encoded and stored in the secrets section of the repo.
+          service_account_key: ${{ secrets.GCR_KEY }}
```
adrianludwin, line 34: Can you say anything about what this service account needs to be able to do? E.g., it must have permission to push to `gcr.io/k8s-staging-multitenancy`? If that's the requirement, how did you test this - by pushing to some different GCR repo?

adrianludwin (commented):
A couple last nits and then I think this looks good to start experimenting with.

/cc @rjbez17 - wdyt?

.github/workflows/image.yaml
rjbez17: Nit: I suggest doing a `+{date}` to the tag as these are automatic builds. A 'real' release should likely only be retagging an intermediate build into the proper version. At least that would be my suggestion, especially since our release include more than just an image build. It also includes docs, kustomization files, and soon to be helm charts.

rjbez17 (commented):
In the future it'd be nice if we split up seperate things into different PRs or at least commits. This PR does a bit more than just multiarch builds.

rjbez17:
lgtm other than the nit around the tagging. I personally think a 'real' release tag should be somewhat manual but open to discussion, but having intermediate builds pushed up will be very helpful.

.github/workflows/image.yaml
RealHarshThakur: I used the `Default compute service account ` as shown in this [blog](http://acaird.github.io/computers/2020/02/11/github-google-container-cloud-run)  and yes, I tested with a different repo.

adrianludwin: Ah. That's bad advice, the default compute service account is terrible and generally shouldn't be used :D But there was no way for you to know that! I take it that the account was in the same GCP project as the repo you used to test?

Maybe just add the comment that we _think_ the requirement for the service accounts it that it can push to the required registry (sorry, "registry" not "repo") but that we're not certain. We'll find out soon enough!

adrianludwin:
> lgtm other than the nit around the tagging. I personally think a 'real' release tag should be somewhat manual but open to discussion, but having intermediate builds pushed up will be very helpful.

+1 to the `+{date}` suggestion for all automatically generated builds for now.

How would you have broken this up? I thought the majority was just "get github actions working" so maybe it's just the wrong title? The actual multiarch builds part seems pretty small.

rjbez17:
> > lgtm other than the nit around the tagging. I personally think a 'real' release tag should be somewhat manual but open to discussion, but having intermediate builds pushed up will be very helpful.
> 
> +1 to the `+{date}` suggestion for all automatically generated builds for now.
> 
> How would you have broken this up? I thought the majority was just "get github actions working" so maybe it's just the wrong title? The actual multiarch builds part seems pretty small.

Yeah so I mean there are three things, one really unrelated, that I see here:

1) Adding github actions to auto-build images based on triggers
2) Actual mutli-arch builds
3) Updating the rbac-proxy version (unless it is needed for multi-arch?) seemed like more of a 'while i'm here' thing.

Not a huge deal but in this case since it isn't a huge amount of code changes here but yeah.

rjbez17:
regardless, lgtm after the +{date} addition

.github/workflows/image.yaml
RealHarshThakur: I added the date in the image tag but I removed the builds being triggered for master branch . Is that okay?

Dockerfile
RealHarshThakur: I think it's okay if we're not explicit about OS and ARCH in the Dockerfile.

.github/workflows/image.yaml
rjbez17: @RealHarshThakur I'm not seeing the date in the tab above, am I missing someething?

RealHarshThakur: @rjbez17 added it [here ](https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/45/commits/2f61d91f2c60221359d8e5e195734bbb235889cc#diff-b4df0a4f0d80f73138c476afbd7aefdac9df339642ddfba323d27c8cbabb92e2R50)

rjbez17:
/lgtm
/approve
/hold 

@adrianludwin want feel free to remove the hold

.github/workflows/image.yaml
adrianludwin: I _think_ that sets the _Docker_ tag so that it includes the date, but `RELEASE_TAG` (which I'm guessing is the _Github_ release tag) still includes only the branch name, correct? E.g., I'd prefer for this to be something like:

```
GITREF="${GITHUB_REF#refs/*/v}"
echo "RELEASE_TAG=${GITREF}-${{ steps.date.outputs.HNC_BUILD_TIMESTAMP }}" >> $GITHUB_ENV
```

In other words, line up the `RELEASE_TAG` and the Docker tags as closely as possible.

@RealHarshThakur if I've misinterpreted please feel free to remove the hold. We can always clean it up later anyway.

adrianludwin (commented):
One last comment, otherwise lgtm. Please feel free to remove the hold yourself (`/hold remove`) if I've misinterpreted, or if you just want to make that change in a followup PR.

RealHarshThakur:
@adrianludwin so we're using branch name + timestamp for image tags. The release tag itself is being used in the metadata during docker build. The reason for not using release tag in the image was because branch names like `master` don't convert well on `TAG="${GITHUB_REF#refs/*/v}`  . Please let me know if that's not okay

adrianludwin:
Ok let's see what it looks like in the GH UI, that should clarify a lot and
will tell us if we need more work.

Thanks for being patient as we went through all these reviews, and congrats
on getting this in! I know that was a lot of work but build systems and
procedures are hard. This is a great contribution!

Issue #47, opened 2021-06-11
Propagation of labels and/or annotations


akazs:
Hello,

We are adopting HNC to our clusters.
However, we usually use labels to manage some information about our namespaces.
Is there any way we can propagate labels and/or annotations to subnamespaces?
If not, do you plan on implementing such a feature?

juv:
We were also looking for this feature while considering HNC vs. Kiosk for our multi-tenancy approach. Unfortunately there is no such option in HNC right now.

For me, this one of those features that makes certain usecases, e.g. aggregate costs of tenants with some kind of cost-aggregation-application, monitoring and alerting or even scraping the API server for all resources of a certain tenant, a lot easier.

adrianludwin:
We've considered it, but we've been lacking:

1. A solid (set of) use cases
2. A concrete technical proposal
3. Enough people willing to help implement the feature

If both of you (@akazs and @juv) could describe your use cases, that would be very helpful. Technically, this _may_ not be terribly technically difficult but it would be easy to implement the "wrong" API by mistake and have to deprecate it and start over. I'd rather not do that if possible.

The big problem is that we don't know _which_ labels (and annotations) should be propagated, and which should not. Possibly worse, if you remove a label from a parent (or move a child to a different parent), we don't know which labels should be _removed_ from the child, because any one of them could have been put there manually by a human, and not HNC. We don't have this problem with regular objects, because all objects propagated by HNC have a label that (if present) tells us that we can modify or delete it. But you can't put a label on a label.

My best guess is to start by adding a list of propagated labels to the cluster-wide `HNCConfiguration` object. E.g., we could say that `foo.com/my-label` is handled by HNC, so you would no longer be able to add it yourself to a namespace, and if a namespace contained that label but its parent did not, HNC would remove it. The only way to set this label would then be in the `HierarchyConfiguration` in the namespace. This would allow us to robustly remove labels (and annotations) from namespaces  when HNC no longer thinks it belongs there.

I'd be willing to do at least some of the implementation work (e.g. the basic propagation and validation) but I'd want some help finishing it up. And I'd like to know if this proposal meets any use cases as well.

Thoughts?

akazs:
We label our namespaces, and use these labels when generating metrics using kubernetes/kube-state-metrics.
This allows us to filter metrics of, say a team, to send alerts or so.
When creating subnamespaces, we would like the same labels to be applied to the newly created subnamespaces.

[...]

PR #48, opened 2021-06-15
[WIP] Propagation of labels

zoetrope:
ref #47

### DIFF
--- a/internal/reconcilers/anchor.go
+++ b/internal/reconcilers/anchor.go
@@ -24,10 +24,8 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	corev1apply "k8s.io/client-go/applyconfigurations/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -37,6 +35,7 @@ import (
 	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
 	"sigs.k8s.io/hierarchical-namespaces/internal/config"
 	"sigs.k8s.io/hierarchical-namespaces/internal/forest"
+	"sigs.k8s.io/hierarchical-namespaces/internal/metadata"
 )
 
 // AnchorReconciler reconciles SubnamespaceAnchor CRs to make sure all the subnamespaces are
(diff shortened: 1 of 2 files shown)


internal/reconcilers/anchor.go
```diff
@@ -106,17 +107,11 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	if inst.Status.State == api.Missing {
-		if err := r.writeNamespace(ctx, log, nm, pnm); err != nil {
-			// Write the "Missing" state to the anchor status if the subnamespace
-			// cannot be created for some reason. Without it, the anchor status will
-			// remain empty by default.
-			if anchorErr := r.writeInstance(ctx, log, inst); anchorErr != nil {
-				log.Error(anchorErr, "while setting anchor state", "state", api.Missing, "reason", err)
-			}
-			return ctrl.Result{}, err
+	if err := r.updateNamespace(ctx, log, nm, pnm); err != nil {
```
adrianludwin, line 110: Note that this change only affects subnamespaces (e.g., child namespaces created via an Anchor), and not _all_ descendant namespaces (e.g., child namespaces that are created normally and then assigned a parent). See [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#basic-subns) for the difference.

As a result, this should probably go in `hierarchy_config.go`, not here.

internal/reconcilers/anchor.go
```diff
@@ -369,18 +364,33 @@ func (r *AnchorReconciler) getNamespace(ctx context.Context, nm string) (*corev1
 	return ns, nil
 }
 
-func (r *AnchorReconciler) writeNamespace(ctx context.Context, log logr.Logger, nm, pnm string) error {
-	inst := &corev1.Namespace{}
+func (r *AnchorReconciler) updateNamespace(ctx context.Context, log logr.Logger, nm, pnm string) error {
+	pnmInst := &corev1.Namespace{}
```
adrianludwin, line 368: Nit: this should be `pnsInst` ("Parent NameSpace Instance").

internal/reconcilers/anchor.go
```diff
@@ -369,18 +364,33 @@ func (r *AnchorReconciler) getNamespace(ctx context.Context, nm string) (*corev1
...
-	// else while this reconciler is running, returning an error will trigger a
-	// retry. The reconciler will set the 'Conflict' state instead of recreating
-	// this namespace. All other transient problems should trigger a retry too.
-	log.Info("Creating subnamespace")
-	if err := r.Create(ctx, inst); err != nil {
-		log.Error(err, "While creating subnamespace")
+	log.Info("Updating subnamespace")
+	if err := r.Patch(ctx, inst, client.Apply, &client.PatchOptions{
+		FieldManager: api.MetaGroup,
+	}); err != nil {
```
adrianludwin, line 392: A couple of things here:
* This was a very important comment, are you sure it's safe to delete?
* Where does the subnamespace get created if you've taken away the "Create" call?
* We use "r.Update" everywhere in HNC, why are you using "r.Patch" here? If this is important, there should be a very clear comment explaining why, otherwise someone will "fix" it in the future.

internal/reconcilers/anchor.go
```diff
@@ -397,15 +407,31 @@ func (r *AnchorReconciler) deleteNamespace(ctx context.Context, log logr.Logger,
 func (r *AnchorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// Maps an subnamespace to its anchor in the parent namespace.
 	nsMapFn := func(obj client.Object) []reconcile.Request {
-		if obj.GetAnnotations()[api.SubnamespaceOf] == "" {
-			return nil
+		result := make([]reconcile.Request, 0)
+		nss := &corev1.NamespaceList{}
+		err := r.List(context.Background(), nss)
```
adrianludwin, line 412: This will result in _every_ namespace being processed _every_ time _any_ namespace is changed, with no timeout. Is this really what you want? Note that the Forest already has all the namespaces in memory.

adrianludwin (commented):
As described [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/47#issuecomment-861427404), there are a bunch of issues we should resolve before we can settle on an implementation, but I took a quick pass and left some comments here. Thanks!

PR #50, opened 2021-06-24
Disable kube-rbac-proxy by default

vishh:
HNC does not expose PII information via `/metrics` endpoint. Disabling employing kube-rbac-proxy by default.

Tested: make test-smoke passes.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -29,13 +29,16 @@ patchesStrategicMerge:
   # Protect the /metrics endpoint by putting it behind auth.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
-- manager_auth_proxy_patch.yaml
+  # DISABLING Auth Proxy by default since HNC does not expose
+  # any PII information via it's metrics.
+  # If this changes in the future, this should be enaled by default. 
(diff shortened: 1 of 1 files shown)


adrianludwin:
/ok-to-test
/cc @rjbez17 

Hey Ryan, the kube-rbac-proxy has been causing us some trouble, e.g. staying on top of the latest version internally at Google for Hierarchy Controller, and also figuring out which multiarch version to use for #45. @vishh (who's a Googler starting to work on HNC) reviewed the metrics and we don't think there's any reason to protect them with the RBAC proxy as it's all pretty innocuous. wdyt?

@vishh, did you deploy this to one of your own clusters and run any e2e tests (e.g. `make test-smoke`)? If so, can you please add a "Tested:" comment to your commit ([example](https://github.com/kubernetes-sigs/hierarchical-namespaces/commit/89b6f66639054ee9bf8f3340b187a227025e8173))? Note that the Prow presubmits don't run e2e tests, only the postsubmits.

vishh:
I can run e2e tests, but I felt a kustomize build was sufficient for this
change.

vishh:
@adrianludwin ran smoke tests and they passed against my GKE cluster.

adrianludwin:
Cool, thanks.
/lgtm
/approve
/hold
/cc @rjbez17 

Ryan, lgty?

PR #51, opened 2021-07-03
Deny requests adding tree labels

RealHarshThakur:
What user can't do anymore :
* Create  a namespace which has a tree label in it
* Add a tree label to a namespace
* Remove a tree label. Controller adds it right back though. We can  prevent the end user from doing this too.  
* Update values of tree label keys. 

Signed-off-by: Harsh Thakur <harshthakur9030@gmail.com>

### DIFF
--- a/internal/namespace/validator.go
+++ b/internal/namespace/validator.go
@@ -3,6 +3,7 @@ package namespaces
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	k8sadm "k8s.io/api/admission/v1"
@@ -86,6 +87,11 @@ func (v *Validator) handle(req *nsRequest) admission.Response {
 		if rsp := v.nameExistsInExternalHierarchy(req); !rsp.Allowed {
 			return rsp
 		}
+
+		if rsp := v.illegalTreeLabel(req); !rsp.Allowed {
+			return rsp
+		}
+
 	case k8sadm.Update:
 		if rsp := v.illegalIncludedNamespaceLabel(req); !rsp.Allowed {
 			return rsp
@@ -96,12 +102,51 @@ func (v *Validator) handle(req *nsRequest) admission.Response {
 		if rsp := v.conflictBetweenParentAndExternalManager(req, ns); !rsp.Allowed {
 			return rsp
 		}
+
+		if rsp := v.illegalTreeLabel(req); !rsp.Allowed {
+			return rsp
+		}
+
 	case k8sadm.Delete:
 		if rsp := v.cannotDeleteSubnamespace(req); !rsp.Allowed {
 			return rsp
 		}
 		if rsp := v.illegalCascadingDeletion(ns); !rsp.Allowed {
 			return rsp
+		}
+	}
+
+	return webhooks.Allow("")
+}
+
+// illegalTreeLabel checks if tree labels are being created or modified
+// by any user or service account since only HNC service account is
+// allowed to do so
+func (v *Validator) illegalTreeLabel(req *nsRequest) admission.Response {
+	msg := "Cannot set or modify tree label %q in namespace %q; these can only be modified by HNC."
+
+	oldLabels := map[string]string{}
+	if req.oldns != nil {
+		oldLabels = req.oldns.Labels
+	}
+	// Ensure the users hasn't added or changed any tree labels
+	for key, val := range req.ns.Labels {
+		if !strings.Contains(key, api.LabelTreeDepthSuffix) {
+			continue
+		}
+
+		// Check if new HNC label tree key isn't being added
+		if oldLabels[key] != val {
+			return webhooks.Deny(metav1.StatusReasonForbidden, fmt.Sprintf(msg, key, req.ns.Name))
+		}
+	}
+
+	for key := range oldLabels {
+		//  Make sure nothing's been deleted
+		if strings.Contains(key, api.LabelTreeDepthSuffix) {
+			if _, ok := req.ns.Labels[key]; !ok {
+				return webhooks.Deny(metav1.StatusReasonForbidden, fmt.Sprintf(msg, key, req.ns.Name))
+			}
 		}
 	}
 
--- a/internal/namespace/validator_test.go
+++ b/internal/namespace/validator_test.go
@@ -339,3 +339,50 @@ func TestIllegalIncludedNamespaceNamespace(t *testing.T) {
 func logResult(t *testing.T, result *metav1.Status) {
 	t.Logf("Got reason %q, code %d, msg %q", result.Reason, result.Code, result.Message)
 }
+func TestIllegalTreeOperations(t *testing.T) {
+	f := foresttest.Create("-")
+	vns := &Validator{Forest: f}
+
+	ns := &corev1.Namespace{}
+	oldNs := &corev1.Namespace{}
+	ns.Name = "a"
+
+	ns.SetLabels(map[string]string{api.LabelTreeDepthSuffix: "1"})
+	// Change tree label value
+	oldNs.SetLabels(map[string]string{api.LabelTreeDepthSuffix: "2"})
+
+	tests := []struct {
+		name    string
+		op      k8sadm.Operation
+		allowed bool
+		oldNs   *corev1.Namespace
+	}{
+		{name: "Creation of namespace with tree label in it", op: k8sadm.Create, allowed: false},
+		{name: "Update namespace with tree label in it", op: k8sadm.Update, allowed: false},
+		{name: "Update value of a tree in existing namespace", op: k8sadm.Update, allowed: false, oldNs: oldNs},
+		{name: "No tree label updates to namespace", op: k8sadm.Update, allowed: true, oldNs: ns},
+	}
+	for _, tc := range tests {
+		g := NewWithT(t)
+
+		var req *nsRequest
+		if tc.op == k8sadm.Create {
+			req = &nsRequest{
+				ns: ns,
+				op: tc.op,
+			}
+		} else {
+			req = &nsRequest{
+				ns:    ns,
+				op:    tc.op,
+				oldns: tc.oldNs,
+			}
+		}
+
+		got := vns.illegalTreeLabel(req)
+
+		logResult(t, got.AdmissionResponse.Result)
+		g.Expect(got.AdmissionResponse.Allowed).Should(Equal(tc.allowed))
+	}
+
+}


internal/validators/namespace.go
```diff
@@ -107,6 +119,36 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+// illegalTreeLabel checks if tree labels are being created or modified
+// by any user or service account since only HNC service account is
+// allowed to do so
+func (v *Namespace) illegalTreeLabel(req *nsRequest) admission.Response {
+	msg := "Only HNC Service account can modify tree labels"
+	treeLabelID := "tree.hnc.x-k8s.io"
+
+	switch req.op {
+	case k8sadm.Update:
+		if !cmp.Equal(req.ns.Labels, req.oldns.Labels) {
```
adrianludwin, line 131: I'd skip this test. It's not actually any faster, and adds a new and unneeded dependency. You could just remove it completely and the code should still work identically.

internal/validators/namespace.go
```diff
@@ -107,6 +119,36 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 	return allow("")
 }
 
+// illegalTreeLabel checks if tree labels are being created or modified
+// by any user or service account since only HNC service account is
+// allowed to do so
+func (v *Namespace) illegalTreeLabel(req *nsRequest) admission.Response {
+	msg := "Only HNC Service account can modify tree labels"
+	treeLabelID := "tree.hnc.x-k8s.io"
```
adrianludwin, line 127: Use `LabelTreeDepthSuffix` (e.g. see [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/be92981add41c07e30946e0c0da5cfa778b5eb4a/internal/reconcilers/hierarchy_config.go#L331)).

internal/validators/namespace.go
```diff
@@ -107,6 +119,36 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+
+	switch req.op {
+	case k8sadm.Update:
```
adrianludwin, line 130: I'd suggest combining the `Create` and `Update` tests, e.g.:

```
oldLabels := map[string]string{}
if req.oldns != nil {
  oldLabels = req.oldns.Labels
}
```

For the `Create` op, `req.oldns` will be nil so every attempt to validate that it's present will fail.

You can also do an early return for delete like:

```
if req.op == k8sadm.Delete {
  return allow("")
}
```

which saves you another level of indentation and lets you skip the switch statement.

@rjbez17 do you agree with this?

adrianludwin (commented):
Can you please add some unit tests in `internal/validators/namespace_test.go` showing both passing and failing cases? Thanks!

RealHarshThakur:
@adrianludwin thanks for the review. Will work on the comments. Meanwhile, can you let me know your thoughts about restricting actions users can still do(check PR description)? I am not sure if they're strictly required but if they are, I'll include it in this PR.

internal/validators/namespace.go
RealHarshThakur: I was following the pattern in other functions with switch statements but your suggestion would make this a whole lot simpler.

adrianludwin: Yeah, the pattern I'm suggesting does exist in other places but it's a bit hidden (e.g. [these three functions](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/be92981add41c07e30946e0c0da5cfa778b5eb4a/internal/validators/object.go#L215).

adrianludwin:
> @adrianludwin thanks for the review. Will work on the comments. Meanwhile, can you let me know your thoughts about restricting actions users can still do(check PR description)? I am not sure if they're strictly required but if they are, I'll include it in this PR.

Ah, sorry, I missed that.

It would definitely be better if we also prevent changes and deletions. With the new structure I'm suggesting, that's also fairly easy - you just do _two_ loops, one over the new labels and one over the old labels. When looping over the new labels you also check the contents so you can disallow changes as well as additions; when looping over the old labels you only need to check for presence (assuming you do it second) to check for deletions.

internal/validators/namespace.go
rjbez17: Yeah I agree

RealHarshThakur:
@adrianludwin  can you PTAL?

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-logr/logr v0.3.0
 	github.com/go-logr/zapr v0.2.0
 	github.com/go-openapi/spec v0.19.5 // indirect
+	github.com/google/go-cmp v0.5.2
```
adrianludwin, line 12: This shouldn't be needed anymore?

go.mod
```diff
@@ -21,6 +22,7 @@ require (
 	k8s.io/apimachinery v0.20.4
 	k8s.io/cli-runtime v0.20.4
 	k8s.io/client-go v0.20.4
+	k8s.io/klog/v2 v2.4.0
```
adrianludwin, line 25: This shouldn't be needed?

internal/validators/namespace.go
```diff
@@ -107,6 +118,46 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+// illegalTreeLabel checks if tree labels are being created or modified
+// by any user or service account since only HNC service account is
+// allowed to do so
+func (v *Namespace) illegalTreeLabel(req *nsRequest) admission.Response {
+	msg := "Only HNC Service account can modify tree labels"
+
+	if req.op == k8sadm.Delete {
+		return allow("")
+	}
+
```
adrianludwin, line 130: Hmm, I had a closer look at the caller and it looks like we only call this function for Create or Update. So you can delete this.

internal/validators/namespace.go
```diff
@@ -107,6 +118,46 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+	oldLabels := map[string]string{}
+	if req.oldns != nil {
+		oldLabels = req.oldns.Labels
+	}
+
+	for key, val := range req.ns.Labels {
```
adrianludwin, line 136: Comment: "Ensure the users hasn't added or changed any tree labels"

internal/validators/namespace.go
```diff
@@ -107,6 +118,46 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+		if oldVal, ok := oldLabels[key]; !ok {
+			if strings.Contains(key, api.LabelTreeDepthSuffix) {
+				return deny(metav1.StatusReasonForbidden, msg)
```
adrianludwin, line 139: It would be nice to say something like:

`"Cannot set or modify tree label %q in namespace %q; these can only be modified by HNC."`

E.g. if this is being called as a result of `kubectl apply -f <directory>`, then the user might not know _which_ namespace in that directory is the problem. And without printing the tree label itself, the user might not know what label we're talking about if they're not HNC experts.

I'd rather not say "Only the HNC service account can do this" because that implies that their mistake was that they're not running as the right service account. But the problem is that they shouldn't be attempting to set these labels at all.

internal/validators/namespace.go
```diff
@@ -107,6 +118,46 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
...
+		return allow("")
+	}
+
+	oldLabels := map[string]string{}
+	if req.oldns != nil {
+		oldLabels = req.oldns.Labels
+	}
+
+	for key, val := range req.ns.Labels {
+		if oldVal, ok := oldLabels[key]; !ok {
```
adrianludwin, line 137: I might restructure this loop a bit. E.g.:

```
if !strings.Contains(...) {
  continue
}
if oldVal, exists := oldLabels[key]; !exists || oldVal != val {
  return deny(...)
}
```

This has less duplication and IMO is a bit easier to read.

internal/validators/namespace_test.go
```diff
@@ -335,3 +335,65 @@ func TestIllegalIncludedNamespaceNamespace(t *testing.T) {
 		})
 	}
 }
+
+func TestNsTreeLabelAddition(t *testing.T) {
```
adrianludwin, line 339: Just `TestTreeLabelAddition` is probably fine here since we're already testing namespace validators so the "Ns" is redundant.

internal/validators/namespace_test.go
```diff
@@ -335,3 +335,65 @@ func TestIllegalIncludedNamespaceNamespace(t *testing.T) {
...
+		}
+		got := vns.illegalTreeLabel(req)
+
+		logResult(t, got.AdmissionResponse.Result)
+		g.Expect(got.AdmissionResponse.Allowed).Should(Equal(tc.fail))
+	}
+
+}
+
+func TestNsUpdateValueTreeLabel(t *testing.T) {
```
adrianludwin, line 371: You should be able to combine this with the last test just by changing the op (and not passing in `oldNS` if the op is `Create`)

internal/validators/namespace_test.go
```diff
@@ -335,3 +335,65 @@ func TestIllegalIncludedNamespaceNamespace(t *testing.T) {
...
+
+	req := &nsRequest{
+		ns:    ns,
+		oldns: oldNs,
+		op:    k8sadm.Update,
+	}
+
+	got := vns.illegalTreeLabel(req)
+
+	g.Expect(got.AdmissionResponse.Allowed).Should(Equal(false))
```
adrianludwin, line 397: We should also test some _success_ cases, including when the new namespace has the _same_ value as the old namespace.

adrianludwin (commented):
Can you please squash your commits as well when you're ready? Thanks!

adrianludwin:
Oh sorry - forgot to add this

/ok-to-test

internal/namespace/validator.go
```diff
@@ -108,6 +119,39 @@ func (v *Validator) handle(req *nsRequest) admission.Response {
...
+		oldLabels = req.oldns.Labels
+	}
+	// Ensure the users hasn't added or changed any tree labels
+	for key, val := range req.ns.Labels {
+		if !strings.Contains(key, api.LabelTreeDepthSuffix) {
+			continue
+		}
+
+		// Check if new HNC label tree key isn't being added
+		if oldVal, ok := oldLabels[key]; !ok || val != oldVal {
```
adrianludwin, line 139: I think you can simplify this to:
```
if oldLabels[key] != val
```
If the key isn't in oldLabels, `oldLabels[key]` will evaluate to the empty string so this case will work for adds as well.

internal/namespace/validator.go
```diff
@@ -108,6 +119,39 @@ func (v *Validator) handle(req *nsRequest) admission.Response {
...
+			return webhooks.Deny(metav1.StatusReasonForbidden, fmt.Sprintf(msg, key, req.ns.Name))
+		}
+	}
+
+	for key := range oldLabels {
```
adrianludwin, line 144: Comment: make sure nothing's been deleted

adrianludwin (commented):
lgtm with two nits (sorry for the delay). Can you please make those changes, and then squash the PR into one commit? Thanks!

RealHarshThakur:
@adrianludwin  sorry, I missed the GH notification. I made the changes but seems like something has changed on the CLA side? I've made commits with the same email before without trouble. Anyway, raised a ticket to check what's wrong

adrianludwin:
Hmm, no idea about the CLA. Sorry for the trouble. I'll review the change, lmk what happens with the CLA stuff.

adrianludwin (commented):
/lgtm
/approve

Issue #25, opened 2021-05-05
HNC: Add a presubmit to ensure `make generate` doesn't produce diffs


yiqigao217:
<a href="https://github.com/yiqigao217"><img src="https://avatars.githubusercontent.com/u/51385444?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [yiqigao217](https://github.com/yiqigao217)**
_Monday Jul 06, 2020 at 18:52 GMT_
_Originally opened as https://github.com/kubernetes-sigs/multi-tenancy/issues/852_

----

In #841 , a change to the auto-generated manifest was not included (patched by #851 later). We should prevent this by adding a presubmit to make sure the `make generate` doesn't produce any diffs.

yiqigao217:
<a href="https://github.com/yiqigao217"><img src="https://avatars.githubusercontent.com/u/51385444?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [yiqigao217](https://github.com/yiqigao217)**
_Monday Jul 06, 2020 at 18:52 GMT_

----

/cc @adrianludwin

yiqigao217:
<a href="https://github.com/fejta-bot"><img src="https://avatars.githubusercontent.com/u/30488982?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [fejta-bot](https://github.com/fejta-bot)**
_Monday Oct 26, 2020 at 20:13 GMT_

----

Issues go stale after 90d of inactivity.
Mark the issue as fresh with `/remove-lifecycle stale`.
Stale issues rot after an additional 30d of inactivity and eventually close.

If this issue is safe to close now please do so with `/close`.

Send feedback to sig-testing, kubernetes/test-infra and/or [fejta](https://github.com/fejta).
/lifecycle stale

yiqigao217:
<a href="https://github.com/adrianludwin"><img src="https://avatars.githubusercontent.com/u/20559844?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [adrianludwin](https://github.com/adrianludwin)**
_Tuesday Oct 27, 2020 at 01:55 GMT_

----

/remove-lifecycle stale

yiqigao217:
[...]

PR #52, opened 2021-07-04
Add presubmit to ensure make generate doesn't produce diffs

xinydev:
Fixes: https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/25  

Keep the same logic as check-fmt 

Maybe we can add [verify-gofmt.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-generated-files.sh), [verify-generated-files.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-generated-files.sh) for these presubmit checks?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,9 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
+check-generate: generate
+	@test -z $(shell git status --untracked-files=no --porcelain)
+
 # Use the version of controller-gen that's checked into vendor/ (see
 # hack/tools.go to see how it got there).
 controller-gen:
--- a/hack/ci-test.sh
+++ b/hack/ci-test.sh
@@ -38,3 +38,6 @@ hack_dir=$(dirname ${BASH_SOURCE})
 echo "Running 'make check-fmt test'"
 cd "$hack_dir/.."
 make check-fmt test
+
+echo "Running 'make check-generate test'"
+make check-generate


adrianludwin (commented):
Nice, thanks! Did you verify that if you deliberately introduced an error, that this test caught it?

/ok-to-test
/lgtm
/approved
/hold
/assign @rjbez17 

If you haven't tested it yet, I'm inclined to merge it anyway but it would then be nice if you could make a fake PR afterwards that tries to break it so we know for certain that it's catching what it's meant to catch.

adrianludwin:
Just glancing at that presubmit failure - is `make check-generate` already failing? That's very possible since we haven't had the check until today. If that's the case, then can you please fix that bug, and don't worry about doing any more testing since it looks like that check works!

xinydev:
@adrianludwin 

>  is `make check-generate` already failing due to an actual error in the repo?

No, It fails becaue `ci-test.sh` download a tar file to the repo folder. I add `--untracked-files=no` to `git status` to fix

> If you haven't tested it yet, I'm inclined to merge it anyway but it would then be nice if you could make a fake PR afterwards that tries to break it so we know for certain that it's catching what it's meant to catch.

I tested it locally.  I will make a fake PR to try to break it for double-check after this PR gets merged

adrianludwin (commented):
/lgtm

Thanks!

adrianludwin:
/hold remove
/approve

@rjbez17 I'm guessing you're out until Wednesday, lmk if you had any concerns with this.

xinydev:
> /hold remove
> /approve
> 
> @rjbez17 I'm guessing you're out until Wednesday, lmk if you had any concerns with this.

maybe `/hold cancel`? 😊

adrianludwin:
Ah man, I am _not_ doing well after my (Canadian) long weekend :D 

/hold cancel

PR #55, opened 2021-07-06
Rename workflow name

RealHarshThakur:
Remove master branch because this would create an image for every commit rather than every release. 

Signed-off-by: Harsh Thakur <harshthakur9030@gmail.com>

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,4 +1,4 @@
-name: release
+name: build
 
 on: 
   create:


adrianludwin:
I actually don't mind creating an image for every commit. Even on a release branch, not every commit _should_ be a release - a release branch is just an _intention_ to create a release in the future, not the release itself (only tags should be interpreted as an intention to create a release at a specific commit).

Does that change the way you feel about this?

adrianludwin:
One thing I'm noticing now is that we do use the word "release" a lot in this flow. Maybe we should just call it a "build" instead of a "release?"

rjbez17:
> One thing I'm noticing now is that we do use the word "release" a lot in this flow. Maybe we should just call it a "build" instead of a "release?"

I agree

RealHarshThakur:
@rjbez17 Just to provide some context from our conversation in [Slack ](https://kubernetes.slack.com/archives/C8E6YA9S7/p1625546793075400). 
TLDR: I think it's better from maintainability perspective to not generate so many images.

adrianludwin:
I'm not overly worried about that - most of the image won't change from release to release, our binary is only 50mb, and we only do a few builds a week. But that's fine.

I would like to change the term "release" to "build" though - can you do that as part of this change?

RealHarshThakur:
@adrianludwin  okay, but do you want the images to be pushed for every commit on master branch too? Or just the `v*`(release) branches?

adrianludwin:
I'd prefer to keep building on the master branch, because otherwise we could break something and not know until we're getting ready to do a release. We can worry about cleaning up the "wasteful" storage artifacts when someone starts complaining about it.

Just for a sense of scale, GCS (which is what GCR uses) charges $0.02/GB/month. The HNC binary is about 50MB, so even if we had a build every day for a year, we'd only be at about 18GB, costing $0.37/mo or $4/yr. In reality, I think we probably average maybe one or two builds a _week_ so I doubt we'd ever come close to costing CNCF $1/yr :)

RealHarshThakur:
@adrianludwin changed it. Please configure the service account credentials before merging this

adrianludwin:
Thanks @RealHarshThakur ! I'm working on getting those secrets configured (it turns out I don't have permission to create a service account myself). I'll let you know when I get it done.

rjbez17:
@adrianludwin thoughts on closing this?

adrianludwin:
It would be great to get this working but I don't have time to look into
it. Any chance you're thinking of doing any HNC work in the near future?

rjbez17:
> It would be great to get this working but I don't have time to look into it. Any chance you're thinking of doing any HNC work in the near future?

I'll take a look at it this weekend. I don't see an issue attached, whats the desired outcome of this? Just a container image built and pushed up to staging for ever push to the main branch?

adrianludwin:
This is to use Github Actions to create official images after every submit - and therefore, for final releases as well - instead of relying on someone (i.e. me) to manually invoke the Cloud Build process whenever it's time to create an official release, and asking people to build nightly images themselves.

rjbez17:
Closing in favor of https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/166

PR #56, opened 2021-07-06
Return non-zero code when pre-submit check fails

xinydev:
We add a pre-submit test to ensure `make generate` doesn't produce diffs by https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/52

This PR is try to break it  for double check

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ fmt:
 # check-fmt: Checks gofmt/go fmt has been ran. gofmt -l lists files whose formatting differs from gofmt's, so it fails if there are unformatted go code.
 check-fmt:
 	@if [ $(shell gofmt -l ${GOFMT_DIRS} | wc -l ) != 0 ]; then \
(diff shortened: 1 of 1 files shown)


adrianludwin:
I'm not sure if we're seeing the correct error message or not. It would be nice if a failure could be clearly identified by something like:

`Error: generated files are out of sync, please run 'make generate' before committing`

adrianludwin:
You may need to rebase. But also please see my comments that I made on the
last PR after it got merged (I think there's a bug I missed).

xinydev:
@adrianludwin  It looks like we solved all the problems?

adrianludwin:
Looks good! Let's just update this PR to add the "exit 1" lines to the makefile - e.g. modify the commit message and the PR title, and then we can merge it. Thanks for this change!

xinydev:
> Looks good! Let's just update this PR to add the "exit 1" lines to the makefile - e.g. modify the commit message and the PR title, and then we can merge it. Thanks for this change!

Thanks for the review!

PR #57, opened 2021-07-06
Make error message for pre-submit check more clearly

xinydev:
from @adrianludwin 's comment https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/56#issuecomment-874822275

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,9 @@ fmt:
 
 # check-fmt: Checks gofmt/go fmt has been ran. gofmt -l lists files whose formatting differs from gofmt's, so it fails if there are unformatted go code.
 check-fmt:
-	@test -z $(shell gofmt -l ${GOFMT_DIRS})
+	@if [ $(shell gofmt -l ${GOFMT_DIRS} | wc -l ) != 0 ]; then \
+		echo "Error: there are unformatted go code, please run 'gofmt' before committing" ; \
+	fi
 
 # Run go vet against code
 vet:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -180,7 +180,9 @@ generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
 check-generate: generate
-	@test -z $(shell git status --untracked-files=no --porcelain)
+	@if [ $(shell git status --untracked-files=no --porcelain | wc -l ) != 0 ]; then \
```
adrianludwin, line 183: While you're at it, maybe fix `check-fmt` too?

adrianludwin (commented):
/ok-to-test

Makefile
xinydev: Thanks for reminding! done

Makefile
```diff
@@ -169,7 +169,9 @@ fmt:
 
 # check-fmt: Checks gofmt/go fmt has been ran. gofmt -l lists files whose formatting differs from gofmt's, so it fails if there are unformatted go code.
 check-fmt:
-	@test -z $(shell gofmt -l ${GOFMT_DIRS})
+	@if [ $(shell gofmt -l ${GOFMT_DIRS} | wc -l ) != 0 ]; then \
+		echo "Error: there are unformatted go code., please run 'gofmt' before committing" ; \
```
adrianludwin, line 173: Minor typo (there shouldn't be a period after "go code").

Also, can you please squash your commits into one? We usually prefer one commit per PR. Thanks!

Makefile
```diff
@@ -180,7 +182,9 @@ generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
 check-generate: generate
-	@test -z $(shell git status --untracked-files=no --porcelain)
+	@if [ $(shell git status --untracked-files=no --porcelain | wc -l ) != 0 ]; then \
+  		echo "Error: generated files are out of sync, please run 'make generate' before committing" ; \
```
adrianludwin, line 186: Oh wait - I know we just merged this but I didn't catch that this won't actually count as an error for `make`. We need to return a non-zero error code, not _just_ print an error message. Sorry for missing that. 

Maybe use something like https://www.gnu.org/software/make/manual/html_node/Make-Control-Functions.html?

adrianludwin: You can probably fix this in the same PR you're using to test to see whether it works or not. AIUI, the presubmit uses the version of the makefile and ci-presubmit.sh (or whatever it's called) that's in the PR, not the one that was previously checked in.

Issue #46, opened 2021-06-03
Create Helm chart as part of the build process


rptaylor:
Hello,

Do you plan to provide a Helm chart for HNC installation?  This would make it easier for operators to manage HNC installations.

Thanks!

adrianludwin:
I'm certainly open to it but I don't really use Helm myself. HNC doesn't really have any options - how would a Helm chart be better than a simple manifest?

Would you be willing to make a PR to add one? If so, I think the way to do it would be similar to how we create the Krew manifest:
* Create the chart file in the `hack/` directory (like the [krew version](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/hack/krew-kubectl-hns.yaml))
* Add a target to the makefile to generate the chart (like [this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/3090e7337b3a19e686cbb720f7f560de5473435e/Makefile#L385))
* Add a stage to the Cloud Build process to upload the chart (like [this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/3090e7337b3a19e686cbb720f7f560de5473435e/cloudbuild.yaml#L78))

rptaylor:
A helm chart can be nice because it bundles all the components together and allows operations  (list, upgrade, uninstall operations)  to be performed on them in aggregate.  Currently with one big YAML file, you can install all the components at once with kubectl apply, but then you are left with a slew of objects and resources in your cluster that can not be cohesively managed.  How would you uninstall HNC? Currently I think there would be no clean way to do this. You could delete the HNC namespace, but you'd still be left with all the un-namespaced objects like CRDs and ClusterRoles.

Also kubectl apply can not apply negative configuration, so if there are some major changes in a new HNC version say from 0.9 to 1.0, and there are old resource definitions in hnc-manager-0.9.yaml  that are not used anymore in hnc-manager-1.0.yaml , they will nevertheless remain in your cluster when you kubectl apply -f hnc-manager-1.0.yaml instead of being removed, adding clutter over time. Once you include something in the HNC YAML file, you will never be able to remove it from the cluster deployments (unless you tell cluster operators to manually kubectl delete things when they upgrade HNC). With Helm, the HNC deployment would be managed so the environment would always be clean, with no leftover resources from previous installations of HNC versions.

Even if there are no configurable HNC options ,
- will that always be the case?
[...]

PR #62, opened 2021-07-15
feat(helm): initial try at making a helm chart

hutchic:
related to https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/46

This is a rough draft of adding a helm chart. Opening as a draft as it will require changes:
- [ ] add a Makefile target that sed's the appVersion and generates `hack/templates/hnc-manager.yaml`
- [ ] adjust cloudbuild.yml to release the helm chart
- [ ] someone should check my usage of helm hooks is sensible
- [ ] open question -- should hnc-manager.yaml have helm templating forced onto it


hack/helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
```
adrianludwin, line 1: Out of interest, is there a canonical version of this file? If so, it might be useful to link to it in a comment.

hack/helm/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.serviceAccount.create -}}
```
adrianludwin, line 1: Why not use the default SA in the `hnc-system` directory? That would eliminate this and (I think) the cluster role binding.

hack/helm/templates/hnc-manager.yaml
```diff
@@ -0,0 +1,721 @@
+apiVersion: v1
```
adrianludwin, line 1: Instead of just checking this in by hand, can you please update the makefile to regenerate it? It's ok to check this in (I think Helm _requires_ that it's checked in, right?), but other than committing it, a developer shouldn't have to do something other than (say) `make manifests`. Otherwise this will quickly drift away from the regular version.

Also, can you please add a comment at the top of this file (automatically inserted, of course) saying something like "Do not edit this file by hand; run 'make manifests' [or whatever] to update."

Thanks!

hack/helm/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
adrianludwin, line 19: I'm not clear on what this is supposed to be. E.g. if we update HNC from v0.8.0 to v0.9.0, should this go from v0.1.0 to v0.2.0?

adrianludwin (commented):
Thanks for this change! Can you please update the docs/releasing.md file as well to describe what we should do when we're releasing a new version (minor or patch)?

hack/helm/templates/_helpers.tpl
hutchic: this is the current latest that gets generated by `helm create chart`

hack/helm/templates/serviceaccount.yaml
hutchic: that would work but my understanding of the hnc docs is we should/must apply the labels first before installing the controller right? If that's the case we need to pull some things out of band specifically
- install the sa / rolebinging
- run the job that adds the label
- now install the controller et al

hack/helm/templates/hnc-manager.yaml
hutchic: that's one of the todo items / why it's a draft PR. Wanted some initial feedback before proceeding down the wrong routes :)

hack/helm/Chart.yaml
hutchic: my understanding which could be incorrect:

- chart version changes when the helm chart itself changes
- app version changes when the application itself changes

so appVersion should be part of the release process to auto bump it. I guess chart version bumps handled in the PR template as process(?)

hack/helm/templates/_helpers.tpl
adrianludwin: Ah got it. Maybe add a "Generated by 'helm create chart'" comment unless that'll be overwritten in the future?

hack/helm/templates/serviceaccount.yaml
adrianludwin: Ah, I see. In that case, can we call it something like `hnc-installer` so that it's clear what this is for? Ditto for the cluster role binding.

Also, if it's essentially a mandatory part of installing HNC, why are we giving users the option to disable it?

As a side note, you actually can apply the labels _after_ installing HNC, but that will mean that there's a period of time during which the namespaces are not actually excluded. Worse, while HNC is starting up, it may forbid changes to the namespaces, so it could take a while to add the labels. So it's definitely _better_ to label them before installing HNC.

hack/helm/values.yaml
```diff
@@ -0,0 +1,16 @@
+nameOverride: ""
+fullnameOverride: ""
```
adrianludwin, line 2: Comment these, i.e. what are the default values ("hnc"?) and what do they control?

hack/helm/values.yaml
```diff
@@ -0,0 +1,16 @@
...
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+hncIgnoreNamespaces:
```
adrianludwin, line 13: `hncExcludeNamespaces` would be better, and you can also point to the right part of the [docs](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#excluding-namespaces-from-hnc) in a comment (as you can tell, I'm pretty big on comments!).

However, simply adding a namespace here will _not_ actually exclude the namespace; you'd also have to add the `----excluded-namespace` option to the HNC deployment. So maybe it's better to hardcode these for now, unless you want to somehow add it to the Deployment in a template?

hack/helm/values.yaml
hutchic: added a comment. This is the current default that get's provided when one does a `helm create` which I'll explain in more detail here and you can tell me if the comment is suitable.

`helm install 123 .` sensible defaults apply and objects get named `hnc-123` which is a combination of Chart.yaml name and the release name
`helm install --set nameOverride=foo 123` objects get named `foo-123`
`helm install --set fullnameOverride=bar 123` objects get named `bar`

hack/helm/values.yaml
hutchic: renamed.

> However, simply adding a namespace here will not actually exclude the namespace

Really? My understanding of the docs was labelling namespaces appropriately would tell the HNC controller to ignore them.

hack/helm/templates/hnc-manager.yaml
hutchic: So we should discuss what to generate and how.

1. Helm standards / common practices for example is to split a file like this up such that there's a `deployment.yaml` file a `serviceaccount.yaml` file and so on. Is it feasible to do this / do we want this?

2. Helm provides a mechanism for namespace creation and targetting. So in theory namespace creation shouldn't be part of the helm chart and anywhere there's a namespace defined it should be a variable `namespace: {{ .Release.Namespace }}`

3. Helm provides a mechanism for label definition again how challenging will this be technically / how badly do we want or need it ie
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "temp.fullname" . }}
  labels:
    {{- include "temp.labels" . | nindent 4 }}
spec:
  {{- if not .Values.autoscaling.enabled }}
  replicas: {{ .Values.replicaCount }}
  {{- end }}
  selector:
    matchLabels:
      {{- include "temp.selectorLabels" . | nindent 6 }}
```

I'm would prefer incremental improvement so a first pass of "an entire static yaml file is generated" with backlog tickets for further helm-ification would be my personal preference but worth discussing before assuming that route is suitable.

adrianludwin: Thanks! And yes, I'm just giving early feedback, I realize this is a draft.

I'd personally prioritize "ease of automation" over "following all the Helm guidelines" for now. I'd rather have something up-to-date and a little nonstandard than a perfect but obsolete chart.

For `deployment.yaml` vs `serviceaccount.yaml` - I think I'd say that if you can generate them from anything in `config/` or from the output of kustomize, then go for it. We can also iterate on that - start as a monolithic file and then use something like the Gatekeeper tooling to split it up.

re namespace - I don't _think_ we use the string `hnc-system` anywhere else, so you could totally add a `sed` command in the makefile to replace it with `{{ .Release.Namespace }}`. Note that there might be strings that _include_ `hnc-system` but that should be pretty easy to check and/or exclude. But I'm totally ok with _not_ doing any of that as well. I think the number of people who would object to `hnc-system` as well as the other names and labels inside that namespace should be vanishingly small, and I'd be very happy with _never_ adding that flexibility unless someone asks for it and/or is willing to do it themselves.

hack/helm/values.yaml
adrianludwin: No, it's a two-step process.
* Why not just let people add the label to the namespace (after all, this is what Kubernetes itself cares about)? Because if we did, anyone who could edit a namespace could exempt themselves from HNC's oversight simply by adding the label, and we don't want to allow that. As a result, if you try to add this label to any normal namespace, HNC will stop you.
* Why not just add the namespace to the `--exclude-namespace` flag, and ask HNC itself to label the namespace on our behalf? Because we don't want HNC to automatically be making changes to critical namespaces like `kube-system`, and users may not want to give it permission to do so anyway.

As a result, we require _both_. The admin of HNC needs to say `--exclude-namespace`, which (now that I think about it) probably should have been called `--allow-namespace-exclusion`, because it gives us _permission_ to add the label to the namespace. And then, someone with edit access to that namespace needs to actually exclude it. Or, if you haven't installed HNC yet, you can do them in the reverse order.

Does that make sense?

hack/helm/templates/hnc-manager.yaml
```diff
@@ -0,0 +1,721 @@
...
+            type: object
+        type: object
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
```
olljanat, line 126: Can you drop `status:` part from CRDs as those make this chart incompatible with https://fleet.rancher.io as it update that part to contains data like:
```yaml
status:
  acceptedNames:
    kind: HierarchyConfiguration
    listKind: HierarchyConfigurationList
    plural: hierarchyconfigurations
    singular: hierarchyconfiguration
  conditions:
  - lastTransitionTime: "2021-09-09T08:15:55Z"
    message: no conflicts found
    reason: NoConflicts
    status: "True"
    type: NamesAccepted
  - lastTransitionTime: "2021-09-09T08:15:55Z"
    message: the initial names have been accepted
    reason: InitialNamesAccepted
    status: "True"
    type: Established
  storedVersions:
  - v1alpha2
 ```
and deployments end up to "Modified" status.

File: hack/helm/templates/hnc-manager.yaml [hunk 8]
```diff
@@ -0,0 +1,721 @@
...
```
olljanat, line 287: Can you drop `status:`  from here too.

hack/helm/templates/hnc-manager.yaml
```diff
@@ -0,0 +1,721 @@
...
```
olljanat, line 347: Can you drop `status:`  from here too.

SydMusavi:
Hi @hutchic 
Are you still working on this issue?

hutchic:
@SydMusavi I am not. It was mentioned somewhere on an issue that if someone wishes to take over to feel free.

rjbez17:
@adrianludwin looks like this has also been dropped. Shall we close?

erikgb:
> @adrianludwin looks like this has also been dropped. Shall we close?

I would like to see a Helm chart to install HNC! Any chance someone could pick this up?

adrianludwin:
The funny thing is, with all the variants we're producing, that just makes
Helm charts _more_ difficult. The real challenge here will be taking the
manifests that are produced by the HNC build process and autogenerating the
charts from them. Gatekeeper's found a way to do this but it would be
nontrivial to adopt it.

I don't want anyone to just build a Helm chart by hand, because that will
quickly become unsupported IMO.

adrianludwin:
I'm going to close this PR, since with the new variant manifests, it's even more infeasible to follow this approach than it was when there was only one. If someone wants to revive the overall idea that's great, but it needs to be automated and based off of the output of `make manifests`.

/close

PR #63, opened 2021-07-20
Update how-to.md

sohaibmaroof:


adrianludwin (commented):
This is actually correct - v0.8 (which is still the most recent version) is in the old repo, while only v0.9 and later (the _next_ version) will be available from this repo. This will make a lot more sense once v0.9 is actually released (probably within a month or so).

Thanks for the suggestion though! I realize it's a bit confusing.

rjbez17:
Should this be abandoned?

PR #67, opened 2021-08-06
Fix wrong port and tree output in quickstart doc

minchao:


docs/user-guide/quickstart.md
```diff
@@ -308,7 +308,7 @@ shell in the client container, confirm that we can access the `s2` service runni
 in namespace `service-2`.
 
 ```bash
-kubectl run s2 -n service-2 --image=nginx --restart=Never --expose --port 8080
+kubectl run s2 -n service-2 --image=nginx --restart=Never --expose --port 80
```
adrianludwin, line 311: I don't think this is correct? The command on line 327 references port 8080, not port 80, and that's the only thing this needs to sync with, correct? Or is there some other reason to change this?

adrianludwin (commented):
Sorry this fell off my radar! One question before approving this.

docs/user-guide/quickstart.md
minchao: I forgot to modify L327, thanks for pointing out. The reason for the modification is that the default port of nginx is 80, unless a customized nginx.conf is provided.

Another alternative is to use the official `nginxinc/nginx-unprivileged` image of nginx, which can keep the existing 8080 port example.

docs/user-guide/quickstart.md
```diff
@@ -308,7 +308,7 @@ shell in the client container, confirm that we can access the `s2` service runni
 in namespace `service-2`.
 
 ```bash
-kubectl run s2 -n service-2 --image=nginx --restart=Never --expose --port 80
+kubectl run s2 -n service-2 --image=nginxinc/nginx-unprivileged --restart=Never --expose --port 8080
```
adrianludwin: Oh, I think I'd use the standard port 80 rather than make this command so much longer. Do you mind making that change (and changing the other 8080s to 80s, of course)?

adrianludwin (commented):
One minor nit, but then can you please squash all the commits into one? Thanks!

minchao:
Sure!

PR #71, opened 2021-08-17
update doc:Use single quotation marks when there is an exclamation point

Abirdcfly:


Abirdcfly:
Thank you for such a detailed document!

adrianludwin (commented):
Great, thanks!

/ok-to-test
/lgtm
/approve

PR #72, opened 2021-08-17
manager should not panic and should ignore wrong Clusterscoped type setting in HNCConfiguration

Abirdcfly:


internal/reconcilers/hnc_config.go
```diff
@@ -186,14 +186,19 @@ func (r *ConfigReconciler) reconcileConfigTypes(inst *api.HNCConfiguration, allR
 		}
 
 		// Look if the resource exists in the API server.
-		gvk, err := GVKFor(gr, allRes)
```
adrianludwin, line 189: Would it be possible to use a custom error type, something like this?

```
gvk, err := GVKFor(gr, allRes)
if err != nil {
  reason := api.ReasonResourceNotFound
  if rle := err.(ResourceLookupError) { // or whatever the syntax is, I forget right now
     reason = rle.reason
  }
  ... everything else pretty obvious ...
}
```

That means we don't have to change the function signature or log/handle "errors" twice.

adrianludwin (commented):
Great change! Do you mind making one API improvement? Please squash all the commits into one before resubmitting. Thanks!

internal/reconcilers/hnc_config.go
Abirdcfly: finish it.

adrianludwin:
Do you mind using a custom Go error? That's the standard way to build this kind of logic (e.g see guide [here](https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go)).

Example:

```
type GVKError struct {
  Reason string
  Msg string
}
func (e *GVKError) Error() string {
  return e.Msg
}
...
// in GVKFor, if not namespaced:
return schema.GroupVersionKind{}, &GVKError{reason: api.ReasonResourceNotNamespaced, msg: fmt.Sprintf("Resource %q is not namespaced", gr))
...
// in caller:
reason := "internal error"
if ge, ok := err.(*GVKError); ok {
  reason = ge.Reason
}
... display the error ...
```

This means that the code calling `GVKFor` doesn't need to know about the format of the error string, and if that error string changes, the caller won't break.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 
/ok-to-test

Issue #22, opened 2021-05-05
windows how to use hns within kubectl


yiqigao217:
<a href="https://github.com/JaneLiuL"><img src="https://avatars.githubusercontent.com/u/56822274?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [JaneLiuL](https://github.com/JaneLiuL)**
_Wednesday Nov 11, 2020 at 07:22 GMT_
_Originally opened as https://github.com/kubernetes-sigs/multi-tenancy/issues/1260_

----

Windows platform, when I download https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-v0.6.0/kubectl-hns_darwin_amd64 and put on the bin folder,  how to use
Now When I try "kubectl hns tree -A" on windows CMD, will fail said "Error: unknown command "hns" for "kubectl""

yiqigao217:
<a href="https://github.com/adrianludwin"><img src="https://avatars.githubusercontent.com/u/20559844?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [adrianludwin](https://github.com/adrianludwin)**
_Wednesday Nov 11, 2020 at 18:46 GMT_

----

We don't pre-build Windows executables so the HNS plugin won't work on
Windows. If you have a Go toolchain you should be able to compile it for
yourself and it *should* work, but since none of the HNC developers have
Windows workstations, we're not able to test it ourselves. If you can build
it and it works, we can add it to our release process.

Will you be able to help with that?
Thanks!

yiqigao217:
<a href="https://github.com/fejta-bot"><img src="https://avatars.githubusercontent.com/u/30488982?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [fejta-bot](https://github.com/fejta-bot)**
_Sunday Mar 14, 2021 at 20:36 GMT_

----

Issues go stale after 90d of inactivity.
Mark the issue as fresh with `/remove-lifecycle stale`.
Stale issues rot after an additional 30d of inactivity and eventually close.

If this issue is safe to close now please do so with `/close`.

Send feedback to sig-contributor-experience at [kubernetes/community](https://github.com/kubernetes/community).
/lifecycle stale

yiqigao217:
[...]

PR #78, opened 2021-09-10
Include Windows CLI to releases

olljanat:
Looks that there have been some discussion about this on kubernetes-sigs/multi-tenancy#1260 and #22 but now PR earlier so here it is.

Storing it with name `kubectl-hns.exe` to any folder on `PATH` environment variable makes `kubectl hns` command working.


rjbez17:
Thanks @olljanat! 

/ok-to-test

/assign @adrianludwin 

This LGTM

adrianludwin (commented):
Sorry for the delay, I was on vacation. Thanks for this!

/lgtm
/approve

PR #80, opened 2021-09-14
Fix hierarchy conflict validation excluding non-propagated objects

joe2far:
- Resolves issue with hierarchy conflict validation
- Excludes objects/resources not being propagated

for example, (behaviour with current logic):
`hnc-parent`  namespace has `kube-root-ca.crt` configmap with applied `propagate.hnc.x-k8s.io/none=true` annotation
when creating a child namespace this will result in the following validation error, even though the parent configmap is not going to be propagated to the child namespace
```
kubectl apply -f child-ns.yaml
kubectl apply -n hnc-child  -f child-ns-hierarchy.yaml
Error from server (Conflict): error when creating "child-ns-hierarchy.yaml": admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "hnc-child": kube-root-ca.crt (/v1, Kind=ConfigMap)
```
with this PR, the parent annotated configmap resource would be excluded from this check and there is no validation error


joe2far:
Adding info on tested scenarios -> 

TESTED: 
```
### with latest hnc version, errors seen for resources with propagate none annotation ->
$ kubectl apply -n hnc-child  -f child-ns-hierarchy.yaml
Error from server (Conflict): error when creating "child-ns-hierarchy.yaml": admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "hnc-child": kube-root-ca.crt (/v1, Kind=ConfigMap)
  * Namespace "hnc-child": istio-ca-root-cert (/v1, Kind=ConfigMap)

------
### now after updating hnc image from this PR
### creation works without issue (and no validation errors)
kubectl apply -n hnc-child  -f child-ns-hierarchy.yaml
hierarchyconfiguration.hnc.x-k8s.io/hierarchy created

### after removing the annotation the conflict validation errors return as expected
kubectl -n hnc-parent annotate cm istio-ca-root-cert propagate.hnc.x-k8s.io/none- # remove annotation
kubectl apply -n hnc-child  -f child-ns-hierarchy.yaml
Error from server (Conflict): error when creating "child-ns-hierarchy.yaml": admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "hnc-child": istio-ca-root-cert (/v1, Kind=ConfigMap)
  ```

File: internal/validators/hierarchy.go [hunk 1]
```diff
@@ -250,11 +251,12 @@ func (v *Hierarchy) getConflictingObjectsOfType(gvk schema.GroupVersionKind, new
...
 	// into the descendants.
 	newAnsSrcObjs := make(map[string]bool)
-	// TODO additionally check if the ancestor source objects obey the
-	//  'shouldPropagateSource()' rules from the reconcilers/object.go. Only
-	//  propagatable ancestor source would cause overwriting conflict.
 	for _, o := range newParent.GetAncestorSourceObjects(gvk, "") {
-		newAnsSrcObjs[o.GetName()] = true
+		// If the user has chosen not to propagate the object to this descendant,
+		// then it should not be included in conflict checks
+		if ok, _ := selectors.ShouldPropagate(o, o.GetLabels()); ok {
```
adrianludwin, line 257: We should never throw away an error - instead, we need to propagate it up to the caller if need be. E.g.:

```
return fmt.Errorf("while determining if this change will conflict with %s %q from namespace %q: %w",
   gvk.Kind, o.GetName(), o.GetNamespace(), err)
```

adrianludwin (commented):
Nice change!

adrianludwin:
Can you please also add a short note about testing to the commit message? E.g.:

```
Tested: got a spurious error on an unpropagated object with this change, which is fixed with this change. Also added a unit test that fails without this change but passes with it.
```

For that second part, that's always a good sanity check if you didn't already try it - just make sure the test fails if you break the feature.

internal/validators/hierarchy.go
joe2far: ShouldPropagate will always return false, even if there is an error, it is essentially replicating the missing conflict logic into hierarchy from object - https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/internal/validators/object.go#L335 and hncconfig https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/internal/validators/hncconfig.go#L169 validations

in these other cases the error is also dropped, as it does not make much to propagate it up to the caller

adrianludwin: Huh. Weird. You make a good point that there's a precedent for throwing the error away so I don't think we should hold up this PR for that reason.

I'm not thrilled about throwing away those error though - they can happen (e.g. if a bug creeps in) and we'll silently discard them. I'll file a separate bug for that.

adrianludwin:
@rjbez17 lmk if you have any concerns about this

/hold cancel

rjbez17:
Sorry I meant to respond earlier. Lgtm as well.

Issue #66, opened 2021-08-03
Add ability to exclude object from propagation based on a common label (fix configmap propagation with Istio)


and-1:
I work with hnc and use configmap propagation to child namespace. When i use hnc in cluster with installed istio, i experience with issue:
`{"level":"info","ts":1627985932.269383,"logger":"validators.Hierarchy","msg":"Denied","ns":"test","op":"CREATE","user":"system:serviceaccount:operators:shell-operator","code":409,"reason":"Conflict","message":"Cannot update hierarchy because it would overwrite the following object(s):\n  * Namespace \"test\": istio-ca-root-cert (/v1, Kind=ConfigMap)\nTo fix this, please rename or remove the conflicting objects first."}`
istio-ca-root-cert is configmap which istio create in all namespaces in cluster. I think hnc has to have extended exclude features based on label of resource or name.

adrianludwin:
That's interesting, I didn't know about this use case. Thanks for submitting it.

You suggested excluding objects based on labels. Does Istio apply a common label to all objects that it creates? If so, I agree that this would be the best option. If not, we could go with names. Do you have the ability to create a PR?

I think this would be a safe addition to make because to exclude an object from propagation, you'd either need to be able to edit the object (in which case, you could just edit the existing exception annotations) or the HNCConfiguration, which is highly privileged. Therefore, this change doesn't open any new security holes. At most, it could _accidentally_ prevent something from being propagated if it _accidentally_ matched the exclusion criteria, but as long as the criteria are written to be tightly scoped, this hopefully won't be a problem.

joe2far:
the `istio-ca-root-cert` does have a specific isitio.io/config label which could be used to exclude it
```
kubectl get cm istio-ca-root-cert -o yaml
...
  labels:
    istio.io/config: "true"
```
however the same issue holds, for the kubernetes created kube-root-ca.crt which does not have a label to easily exclude it on
```
kubectl get cm kube-root-ca.crt -o yaml
...
   which has no labels
```
given above I would prefer to have solution to exclude based upon specific resource names, this would solve both of the above examples and would follow a similar pattern to RBAC rules which allows api_group, api_version, resource/kind or resourceNames to be specified.

adrianludwin:
Hey @joe2far sorry I didn't respond earlier but I've been on vacation since Labour Day.

[...]

PR #81, opened 2021-09-15
Add Excluded Resource Names to HNC config

joe2far:
PR add built-in exception/exclusion list, of configmap names that are excluded from propagation

It was proposed that add a listed of excluded resource names to hnc-config as below:
But this may be something that is added later when there are other clearer usecases
```
apiVersion: hnc.x-k8s.io/v1alpha2
kind: HNCConfiguration
metadata:
  name: config
spec:
  resources:
  - mode: Propagate
    resource: secrets
  - mode: Propagate
    resource: configmaps
    excludedNames: ["istio-ca-root-cert", "kube-root-ca.crt"]  
```
Any resource matching the listed names will fail shouldPropogate and not be propagated by hnc

this give solution to istio / kubernetes configmap resources which are auto generated in every namespace and cannot be propagated due to conflicts, it may help for other similarly created resources which need to be excluded

note: helps to resolve https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/66
 but by explicit exclusions of resource names (rather than using labels) so as to address `kube-root-ca.crt` as well `istio-ca-root-cert` configmaps

### DIFF
--- a/internal/pkg/selectors/selectors.go
+++ b/internal/pkg/selectors/selectors.go
@@ -29,6 +29,9 @@ func ShouldPropagate(inst *unstructured.Unstructured, nsLabels labels.Set) (bool
 	if none, err := GetNoneSelector(inst); err != nil || none {
 		return false, err
 	}
+	if excluded, err := isExcluded(inst); excluded {
+		return false, err
+	}
 	return true, nil
 }
 
@@ -153,4 +156,21 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
 
 	}
 	return noneSelector, nil
+}
+
+// cmExclusions are known (istio and kube-root) CA configmap which are excluded from propagation
+var cmExclusions = []string{"istio-ca-root-cert", "kube-root-ca.crt"}
+
+// isExcluded returns true to indicate that this object is excluded from being propagated
+func isExcluded(inst *unstructured.Unstructured) (bool, error) {
+	name := inst.GetName()
+	kind := inst.GetKind()
+	group := inst.GroupVersionKind().Group
+
+	for _, excludedResourceName := range cmExclusions {
+		if group == "" && kind == "ConfigMap" && name == excludedResourceName {
+			return true, nil
+		}
+	}
+	return false, nil
 }
--- a/internal/reconcilers/object_test.go
+++ b/internal/reconcilers/object_test.go
@@ -390,6 +390,19 @@ var _ = Describe("Basic propagation", func() {
 		Expect(objectInheritedFrom(ctx, "configmaps", barName, "foo-config")).Should(Equal(fooName))
 	})
 
+	It("should not propagate builtin exclusions", func() {
+		setParent(ctx, barName, fooName)
+		makeObject(ctx, "configmaps", fooName, "istio-ca-root-cert")
+		makeObject(ctx, "configmaps", fooName, "kube-root-ca.crt")
+		makeObject(ctx, "configmaps", fooName, "gets-propagated")
+		addToHNCConfig(ctx, "", "configmaps", api.Propagate)
+
+		// We expect normal configmaps to be propagated, but builtin exclusions not to be.
+		Eventually(hasObject(ctx, "configmaps", barName, "gets-propagated")).Should(BeTrue())
+		Eventually(hasObject(ctx, "configmaps", barName, "istio-ca-root-cert")).Should(BeFalse())
+		Eventually(hasObject(ctx, "configmaps", barName, "kube-root-ca.crt")).Should(BeFalse())
+	})
+
 	It("should be removed if the hierarchy changes", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
--- a/docs/user-guide/concepts.md
+++ b/docs/user-guide/concepts.md
@@ -348,6 +348,19 @@ overwritten by your actions. You can then rewrite the exception to safely
 exclude those objects, or else delete the conflicting objects to allow them to
 be replaced.
 
+#### Built-in exceptions
+
+There are some built-in exceptions to prevent certain known (auto-generated)
+objects from being propagated by HNC.
+
+If ConfigMaps propagation is enabled, any ConfigMaps named `istio-ca-root-cert`
+or `kube-root-ca.crt` will not be propagated. These are auto-created in new
+namespaces by Istio and Kubernetes respectively. As they are auto-generated,
+adding annotations is not possible and HNC will by default exclude them.
+
+Similarly, Kubernetes ServiceAccount Secrets will also by default be excluded
+from propagation.
+
 <a name="admin"/>
 
 ## Administration


internal/pkg/selectors/selectors.go
```diff
@@ -154,3 +157,28 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
 	}
 	return noneSelector, nil
 }
+
+var resourceExclusions = []string{"istio-ca-root-cert", "kube-root-ca.crt"}
+
+// CheckForExclusion returns true to indicate that this object is excluded from being propagated
```
adrianludwin, line 163: Maybe `checkForExclusions` (plural) is `isExcluded`? Also make it private until we have some reason to want to export this function.

adrianludwin (commented):
I've left some comments on #66 , I think it might actually be preferable to hardcode these exclusions since K8s (obviously) and Istio are so fundamental that it makes sense to support them out-of-the-box. We can always switch this to a suitable default on the `HNCConfiguration` later if need be.

Maybe let's hold off on changes to this PR for now, but once we agree on the approach, can we please add a unit and maybe end-to-end test? And let's also update the documentation in the same PR to document this new behaviour.

/ok-to-test
/cc @rjbez17 

Ryan this is just FYI for now until we agree on the overall approach.

internal/pkg/selectors/selectors.go
joe2far: have now updated this method to isExcluded()

joe2far:
> I've left some comments on #66 , I think it might actually be preferable to hardcode these exclusions since K8s (obviously) and Istio are so fundamental that it makes sense to support them out-of-the-box. We can always switch this to a suitable default on the `HNCConfiguration` later if need be.
> 
> Maybe let's hold off on changes to this PR for now, but once we agree on the approach, can we please add a unit and maybe end-to-end test? And let's also update the documentation in the same PR to document this new behaviour.
> 
> /ok-to-test
> /cc @rjbez17
> 
> Ryan this is just FYI for now until we agree on the overall approach.

- I have now updated the PR to just contain the these hardcoded exclusions for CA configmaps (assuming this is sufficent for now, agree we can add to config later if needed but it is better for these to be excluded by default)
- Also have added unit test to validate that these CA configmaps are now not propagated

internal/reconcilers/object_test.go
```diff
@@ -390,6 +390,17 @@ var _ = Describe("Basic propagation", func() {
...
 
+	It("should not be copied to descendents when source object is excluded", func() {
+		setParent(ctx, barName, fooName)
+		makeObject(ctx, "configmaps", fooName, "istio-ca-root-cert")
+		makeObject(ctx, "configmaps", fooName, "kube-root-ca.crt")
+		addToHNCConfig(ctx, "", "configmaps", api.Propagate)
+
+		// excluded ca cert configmaps should not be propagated from foo to bar.
+		Eventually(hasObject(ctx, "configmaps", barName, "istio-ca-root-cert")).Should(BeFalse())
+		Eventually(hasObject(ctx, "configmaps", barName, "kube-root-ca.crt")).Should(BeFalse())
```
adrianludwin, line 401: This could give a false positive even if the object _were_ propagated, because propagation takes time and this line could be executed before it completes.

A slightly better approach is to have an object that _is_ propagated, wait for that to be propagated _first_, and then use `Expect` instead of `Eventually` (since we're really not actually polling). You could also use `Consistently` which always polls _until_ the timeout completes, but TBH that's probably overkill here.

Whichever you pick, it's probably worth commenting too though since this is often not obvious to future devs!

adrianludwin (commented):
This looks almost ready to me, with a few last caveats:
1. Can you please add a comment to the commit message (not just this PR) about how you tested it? E.g. "Added unit tests, plus confirmed that I don't see conflicts on K8s 1.22 or with Istio anymore."
2. Can you please document this new behaviour? Maybe add a subsection in the [exceptions](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#exceptions-and-propagation-control) doc called "Built-in exceptions" where you discuss these two exceptions as well as the fact that K8s service account secrets are also not propagated.

Otherwise, this lgtm. Thanks!

joe2far:
> This looks almost ready to me, with a few last caveats:
> 
> 1. Can you please add a comment to the commit message (not just this PR) about how you tested it? E.g. "Added unit tests, plus confirmed that I don't see conflicts on K8s 1.22 or with Istio anymore." **[UPDATE]:** Sorry I just saw that you did update the commit message with the unit tests (sorry I missed that). It might still be worth updating it with your description of the e2e test as well.
> 2. Can you please document this new behaviour? Maybe add a subsection in the [exceptions](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#exceptions-and-propagation-control) doc called "Built-in exceptions" where you discuss these two exceptions as well as the fact that K8s service account secrets are also not propagated.
> 
> Otherwise, this lgtm. Thanks!

Documentation now updated to outline built-in propagation exceptions

internal/reconcilers/object_test.go
joe2far: have added a propagated resource to this test and update these checks to `Expect` as suggested

internal/pkg/selectors/selectors.go
```diff
@@ -154,3 +157,19 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
...
+// cmExclusions are known (istio and kube-root) CA configmap which are excluded from propagation
+var cmExclusions = []string{"istio-ca-root-cert", "kube-root-ca.crt"}
+
+// isExcluded returns true to indicate that this object is excluded from being propagated
+func isExcluded(inst *unstructured.Unstructured) (bool, error) {
+	name := inst.GetName()
+	kind := inst.GetKind()
+
+	for _, excludedResourceName := range cmExclusions {
+		if name == excludedResourceName && kind == "ConfigMap" {
```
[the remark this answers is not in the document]
joe2far: have now added - group == ""

internal/reconcilers/object_test.go
joe2far: using `Expect` this test may periodically fail ref:  https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_hierarchical-namespaces/81/pull-hnc-test/1440412972798185472
so have switched back to `Eventually` but have added a configmap that is propagated to the test

rjbez17:
Hmm, I'm really on the fence with this one. The Istio issue has plagued me for a while but I generally dislike adding some hidden "magic" for specific things out of my control. I would generally prefer these being a default in an overall HNCConfiguration, but I'm not sure I'm concerned enough to hold up this PR.

adrianludwin:
> Hmm, I'm really on the fence with this one. The Istio issue has plagued me for a while but I generally dislike adding some hidden "magic" for specific things out of my control. I would generally prefer these being a default in an overall HNCConfiguration, but I'm not sure I'm concerned enough to hold up this PR.

FTR - Ryan and I synced on this offline and we agreed that this should go forward, and we shouldn't try to add a new API until we have some more use cases (e.g. to settle on name vs labels).

internal/reconcilers/object_test.go
```diff
@@ -390,6 +390,19 @@ var _ = Describe("Basic propagation", func() {
 		Expect(objectInheritedFrom(ctx, "configmaps", barName, "foo-config")).Should(Equal(fooName))
 	})
 
+	It("should not be copied to descendents when source object is excluded", func() {
```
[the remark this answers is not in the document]
joe2far: have now updated the test describe

internal/reconcilers/object_test.go
```diff
@@ -390,6 +390,19 @@ var _ = Describe("Basic propagation", func() {
...
+		setParent(ctx, barName, fooName)
+		makeObject(ctx, "configmaps", fooName, "istio-ca-root-cert")
+		makeObject(ctx, "configmaps", fooName, "kube-root-ca.crt")
+		makeObject(ctx, "configmaps", fooName, "gets-propagated")
+		addToHNCConfig(ctx, "", "configmaps", api.Propagate)
+
+		// excluded ca cert configmaps should not be propagated from foo to bar.
+		Expect(objectInheritedFrom(ctx, "configmaps", barName, "gets-propagated")).Should(Equal(fooName))
+		Eventually(hasObject(ctx, "configmaps", barName, "istio-ca-root-cert")).Should(BeFalse())
+		Eventually(hasObject(ctx, "configmaps", barName, "kube-root-ca.crt")).Should(BeFalse())
```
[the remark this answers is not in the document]
joe2far: Have now updated this test and comment

adrianludwin:
Thanks for this change!

/approve
/lgtm

PR #83, opened 2021-09-26
Use controller-runtime 0.10.1 and Go 16 

adrianludwin:
Also updates controller-tools to v0.7.0, which required some changes to
the makefile since some options have been removed.

Tested: unit and smoke tests


adrianludwin:
Whoops didn't mean to push this directly to the repo

PR #84, opened 2021-09-26
Use controller-runtime 0.10.1 and Go 16

adrianludwin:
Also updates controller-tools to v0.7.0, which required some changes to
the makefile since some options have been removed.

Tested: unit and smoke tests


adrianludwin:
This is currently blocked by https://github.com/kubernetes/test-infra/pull/23754.

adrianludwin:
@rjbez17 looks like we're good to go if you are

PR #85, opened 2021-09-27
Upgrade to Go 1.17

adrianludwin:
Tested: unit tests; partial smoke test (my workstation is seeing random
timeouts).


adrianludwin:
I expect this to be blocked by https://github.com/kubernetes/test-infra/pull/23756 due to the change in `go fmt` from 1.16 to 1.17.

adrianludwin:
As expected, this fails when built by Go 1.16.

Issue #87, opened 2021-09-30
HNC: excluded-namespace and included-namespace


santinoncs:
Hi,

I want to use HNC but only in a few specific namespaces.  Because of that, it is far more difficult to exclude-namespaces and include few than the other way. 
Also it would be great if you can use regexp
Do you think is feasible ? it makes sense? 
this would be like

        - args:
          - --webhook-server-port=9443
          - --metrics-addr=:8080
          - --max-reconciles=10
          - --apiserver-qps-throttle=50
          - --enable-internal-cert-management
          - --include-namespaces="tests-.*"

adrianludwin:
Can you explain what you mean by "use HNC... in a few specific namespaces?"
By and large, if you don't put any HNC configs into a namespace, HNC will
ignore it, *unless* something goes terribly wrong with HNC.
--exclude-namespaces exists to protect the most critical namespaces on your
cluster from HNC going nuts; it's not really a user-visible feature, so
much as it's a way to make sure that no matter how badly HNC behaves, the
rest of the cluster is operating well enough so that (say) you can delete
HNC.

It's also important to understand that --exclude-namespaces on its own as
actually just a backup feature that operates on *non* excluded namespaces.
Let me explain: HNC protects critical system namespaces (like kube-system)
by having its webhook include a namespace selector
<https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector>
that excludes all namespaces with the label hnc.x-k8s.io/excluded-namespace
(unfortunately, webhooks don't let you exclude namespaces by name). But
this means that anyone who can edit a namespace could just *add* this label
to their own namespace, and HNC would start to ignore it. Since HNC is part
of the security system of your entire cluster, we shouldn't allow this, and
so if HNC sees that someone's added the label to any namespace, it
automatically removes it - *unless* that namespace is part of
--exclude-namespaces.

With that said, I can see a good reason to implement this feature - if you
have very large clusters, and you don't trust HNC yet (a very reasonable
position), but want to use it in some areas. If that matches what you're
trying to do, then I'd totally support it if someone wanted to make a PR,
and I'll help review it. Here's how I'd approach this problem:

   - Modify the webhook (here
   <https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/ba15870be8ba46b259891c1cd55634e7e438d62c/config/webhook/webhook_patch.yaml>)
[...]

PR #89, opened 2021-10-07
Add --include-namespaces-regex features

santinoncs:
In order to just specify few namespaces to be managed by HNC we implemented this logic. So in the case you want to exclude everything except specific namespaces, you can set this parameter

> --included-namespace-regex

Check this out for more info

https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/87


adrianludwin:
This is a great change, thanks so much!

/lgtm
/approve

santinoncs:
> This is a great change, thanks so much!
> 
> 
> 
> /lgtm
> 
> /approve

Thanks for your help

PR #93, opened 2021-10-12
Improve usability of unmanaged namespaces

adrianludwin:
This change makes it a nondestructive operation to switch a namespace
from managed to unmanaged and back, by _not_ deleting the HC or anchors
while a namespace is unmanaged. It also fixes one serious bug
(namespaces could not be changed to roots because the unnamed namespace
didn't match the regex) and adds a lot more details in the error
messages. Finally, it allows users to fix broken hierarchies by removing
a formerly-managed parent.

I also changed the terminology from "excluded" to "unmanaged"
throughout.

Testing: This is hard to test automatically since it requires constant
redeployments of HNC. So I did extensive manual testing instead:

* Try to arrange hierarchies with unmanaged namespace, both excluded
(e.g hnc-system) and ones that don't match the regex. Verify that the
error messages clearly indicate the correct problem.
* Create a legal hierarchy and then break it by changing the regex.
Verify that I can fix it without restarting HNC.
* Create a legal hierarchy, break it with the regex, and then restore it
with the regex. Verify that the original hierarchy comes back.
* Verify that anchors and hierarchyconfigs lose their finalizers when
their namespaces become unmanaged, and get them back if they become
managed again.

### DIFF
--- a/internal/anchor/reconciler.go
+++ b/internal/anchor/reconciler.go
@@ -75,23 +75,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
-	// Always delete anchor (and any other HNC CRs) in the excluded namespaces and
-	// early exit.
-	if !config.IsNamespaceIncluded(pnm) {
-		// Since the anchors in the excluded namespaces are never synced by HNC,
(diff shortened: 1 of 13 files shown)


adrianludwin:
/assign @rjbez17 
cc @santinoncs

santinoncs:
LGTM

PR #119, opened 2022-01-12
Simplify condition handling

adrianludwin:
While implementing managed labels, I found two problems with our
condition handling:

1. They were using old terminology like "Critical Condition" instead of
the newer "ActivitiesHalted".
2. Instead of identifying problems like cycles right away, we delayed
their identification until the end of the syncing process. This means
that code that depends on knowing about cycles - like setting tree
labels - needed to be at the end of the sync process instead of in the
middle where they belonged (and couldn't set conditions themselves).

This change fixes both problems.

Tested: all existing unit/integ tests; no change in functionality.

### DIFF
--- a/internal/forest/namespaceconditions.go
+++ b/internal/forest/namespaceconditions.go
@@ -1,36 +1,47 @@
 package forest
 
 import (
+	"fmt"
+
 	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
 )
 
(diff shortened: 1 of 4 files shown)


adrianludwin:
/assign @rjbez17 

This is the first of three PRs to implement #47

adrianludwin:
I forget if I can self-approve or not...

/lgtm

adrianludwin:
Nope :) Will wait for Ryan

Issue #104, opened 2021-10-19
Let HNC disable propagation Rancher-propagated objects


ichmachnixichgucknur:
Hi,
I am new to the HNC and wondering how to disable the default resources (roles and rolebinding)?
I'd like to use the HNC just to propagate networkpolicies and resource limits to the child namespaces.

ichmachnixichgucknur:
We are using Rancher. Rancher does his own magic on permissions, roles etc. .
I tried some stuff with the HNC and it fits perfectly to distribute networkpolicies and resource limits, but we get into problems with roles and rolebindings. So if I could disable the defaults, that would be really great.
I am not a go dev - but if you are willing to guide me ... I can try to.

ichmachnixichgucknur:
yes, exactly. Rancher is organized in projects and projects inherit namespaces. that's just done using annotations. permissions are set on the projects and applied by rancher to the namespaces.
this leads to some rolebindings that are the same all over the clusters.

using the HNC I get these errors, because it wants to delete an account that is created by rancher:
`{"level":"info","ts":1634650448.2220306,"logger":"validators.Object","msg":"Denied","resource":{"group":"rbac.authorization.k8s.io","version":"v1","resource":"rolebindings"},"ns":"netpoltest-101","nm":"rb-pmsrakgptq","op":"DELETE","user":"system:serviceaccount:cattle-system:kontainer-engine","code":403,"reason":"Forbidden","message":"Cannot delete object propagated from namespace \"my-root-ns\""}`

ichmachnixichgucknur:
We will try and come back to you.

ichmachnixichgucknur:
> * We'll need to find a way to ensure that the _initial_ HNCConfiguration object has Roles and Rolebindings included, since I want _something_ to work out of the box. This isn't a big deal, but will need to be coded and tested.

Yes, I did not find the right part to do it. The easiest would be to deploy a hnc configuration including this types if no config exists.

> * This won't help cases where you want to propagate _some_ RBAC permissions. In that case, you'd still need to do the per-object-level filtering I mentioned.

Correct.
 
> So since we need per-object-level filtering anyway, maybe we can start by just building that?

Do we have two issues here? One thing is to be more flexible having the roles and rolebindings as flexible as the other resources are. That would work with the changes we did, currently missing the default role and rolebinding (so that tests will run, too).
The second thing would be the per per-object filtering - maybe as a cli-parameter.

[...]

PR #123, opened 2022-01-18
Add cli argument `--disable-enforced-types`

nobbs:
This kinda implements the request made in #104, as it makes it possible to disable the default enforced types `roles` and `rolebindings` with an additional cli argument `--disable-enforced-types` that can be used with the hnc-manager. The old behaviour (using enforcedTypes) is the default.

Tested: added new e2e tests and checked them against a vanilla k8s cluster. No issues found.

---

I'll try to implement the per-object-level filtering proposed by @adrianludwin in https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/104#issuecomment-946865077 in another PR soon.

If there are any issues with this PR, please let me know. This being my first contribution to a k8s / cncf project, I've probably missed something ;)

### DIFF
--- a/internal/hncconfig/reconciler.go
+++ b/internal/hncconfig/reconciler.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
+	"sigs.k8s.io/hierarchical-namespaces/internal/config"
 	"sigs.k8s.io/hierarchical-namespaces/internal/crd"
 	"sigs.k8s.io/hierarchical-namespaces/internal/forest"
 	"sigs.k8s.io/hierarchical-namespaces/internal/objects"
@@ -155,9 +156,12 @@ func (r *Reconciler) reconcileTypes(inst *api.HNCConfiguration) error {
 	// Overwrite the type set each time. Initialize them with the enforced types.
 	r.activeGVKMode = gr2gvkMode{}
 	r.activeGR = gvk2gr{}
-	if err := r.ensureEnforcedTypes(inst, allRes); err != nil {
-		// Early exit if any enforced types are not found for some reason to retry.
-		return err
+
+	if !config.EnforcedTypesDisabled {
+		if err := r.ensureEnforcedTypes(inst, allRes); err != nil {
+			// Early exit if any enforced types are not found for some reason to retry.
+			return err
+		}
 	}
 
 	// Add all valid configurations from user-configured types.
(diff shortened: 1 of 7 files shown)


internal/hncconfig/reconciler.go
```diff
@@ -155,9 +156,12 @@ func (r *Reconciler) reconcileTypes(inst *api.HNCConfiguration) error {
 	// Overwrite the type set each time. Initialize them with the enforced types.
 	r.activeGVKMode = gr2gvkMode{}
 	r.activeGR = gvk2gr{}
-	if err := r.ensureEnforcedTypes(inst, allRes); err != nil {
-		// Early exit if any enforced types are not found for some reason to retry.
-		return err
+
+	if !config.EnforcedTypesDisabled {
+		if err := r.ensureEnforcedTypes(inst, allRes); err != nil {
```
adrianludwin, line 161: This should be an early return from inside `ensureEnforcedTypes` IMO. It's better not to clutter up the callers of the function if it can reasonable be handled _inside_ the function.

internal/hncconfig/reconciler.go
```diff
@@ -200,7 +204,7 @@ func (r *Reconciler) reconcileConfigTypes(inst *api.HNCConfiguration, allRes []*
 			log := r.Log.WithValues("resource", gr, "appliedMode", gvkMode.mode)
 			msg := ""
 			// Set a different message if the type is enforced by HNC.
-			if api.IsEnforcedType(rsc) {
+			if !config.EnforcedTypesDisabled && api.IsEnforcedType(rsc) {
```
adrianludwin, line 207: Ditto, we should do this check inside `api.IsEnforcedType`. That may mean moving the global var from `config` to `api` instead since I think `config` imports `api` but not vice versa.

pkg/testutils/testutils.go
```diff
@@ -382,6 +382,26 @@ func RecoverHNC() {
 	cleanupNamespaces(a, b)
 }
 
+func AddDisableEnforcedTypesArg() {
+	// If hncRecoverPath is not set, we should not get here, but if we do, return early and - propably - fail the tests
+	// as the changes done in this prep function require the original manifest file to recover a "normal" state.
+	if hncRecoverPath == "" {
+		return
+	}
+
```
adrianludwin, line 391: As long as the test calls `CheckHNCPath()` before calling this, we shouldn't need these lines. There are many cases where this _isn't_ set and we don't want the tests to fail.

adrianludwin (commented):
I haven't finished reviewing the tests but thought I'd give some early feedback. PTAL! I'll try to get back to those tests tomorrow. Thanks for the change!

adrianludwin:
Hmm, do we still need this if we have #125 ? I far prefer #125 to this change.

adrianludwin:
As discussed on #125 , we'll go with that solution instead of this one.

/close

PR #125, opened 2022-01-19
Add per-object-level filtering based on labels (excluding Rancher-generated resources)

nobbs:
The set of rules is hardcoded as of now, as proposed in #104. The current rules are Rancher-specific, as Rancher does some propagation of roles, rolebindings and serviceaccounts on its own via its "projects" (used to group namespaces). Label equality is checked against both key and value - no regex support for now.

Tested: addded a new test to check whether roles propagation is prevented if the label is matched; also checked, that this prevents propagation only for the specified groups / kinds.

---

I'm not really satisfied with hardcoding these, I personally would prefer a configurable solution. Unfortunately, I'm not sure how to best implement that.

### DIFF
--- a/internal/integtest/helpers.go
+++ b/internal/integtest/helpers.go
@@ -300,9 +300,9 @@ func MakeObject(ctx context.Context, resource string, nsName, name string) {
 	createdObjects = append(createdObjects, inst)
 }
 
-// MakeObjectWithAnnotation creates an empty object with annotation given kind in a specific
+// MakeObjectWithAnnotations creates an empty object with annotation given kind in a specific
 // namespace. The kind and its corresponding GVK should be included in the GVKs map.
-func MakeObjectWithAnnotation(ctx context.Context, resource string, nsName,
+func MakeObjectWithAnnotations(ctx context.Context, resource string, nsName,
 	name string, a map[string]string) {
 	inst := &unstructured.Unstructured{}
 	inst.SetGroupVersionKind(GVKs[resource])
@@ -313,9 +313,22 @@ func MakeObjectWithAnnotation(ctx context.Context, resource string, nsName,
 	createdObjects = append(createdObjects, inst)
 }
 
-// UpdateObjectWithAnnotation gets an object given it's kind, nsName and name, adds the annotation
+// MakeObjectWithLabels creates an empty object with label given kind in a specific
+// namespace. The kind and its corresponding GVK should be included in the GVKs map.
+func MakeObjectWithLabels(ctx context.Context, resource string, nsName,
+	name string, l map[string]string) {
+	inst := &unstructured.Unstructured{}
+	inst.SetGroupVersionKind(GVKs[resource])
+	inst.SetNamespace(nsName)
+	inst.SetName(name)
+	inst.SetLabels(l)
+	ExpectWithOffset(1, K8sClient.Create(ctx, inst)).Should(Succeed())
+	createdObjects = append(createdObjects, inst)
+}
+
+// UpdateObjectWithAnnotations gets an object given it's kind, nsName and name, adds the annotation
 // and updates this object
-func UpdateObjectWithAnnotation(ctx context.Context, resource string, nsName,
+func UpdateObjectWithAnnotations(ctx context.Context, resource string, nsName,
 	name string, a map[string]string) error {
 	nnm := types.NamespacedName{Namespace: nsName, Name: name}
 	inst := &unstructured.Unstructured{}
--- a/internal/objects/reconciler_test.go
+++ b/internal/objects/reconciler_test.go
@@ -142,7 +142,7 @@ var _ = Describe("Exceptions", func() {
 				tc.treeSelector = ReplaceStrings(tc.treeSelector, names)
 
 				// Create a Role with the selector and treeSelector annotation
-				MakeObjectWithAnnotation(ctx, api.RoleResource, names[p], "testrole", map[string]string{
+				MakeObjectWithAnnotations(ctx, api.RoleResource, names[p], "testrole", map[string]string{
 					api.AnnotationSelector:     tc.selector,
 					api.AnnotationTreeSelector: tc.treeSelector,
 					api.AnnotationNoneSelector: tc.noneSelector,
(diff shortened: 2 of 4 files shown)


internal/selectors/selectors.go
```diff
@@ -158,19 +158,75 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
...
+			allLabelsFound := true
+			for excludedKey, excludedValue := range res.Labels {
+				labelValue, ok := labels[excludedKey]
+				if !ok || excludedValue != labelValue {
+					allLabelsFound = false
+				}
+			}
+			if allLabelsFound {
+				return true, nil
+			}
```
adrianludwin, line 227: Does this need to be so complex? Right now, all of the proposed exclusions only look at a single label. If all known cases would be fine with a single key-value pair, let's simplify it to that and only add this complexity when we need it?

internal/selectors/selectors.go
```diff
@@ -158,19 +158,75 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
...
+var ExclusionsByName = []ExclusionByNameSpec{
+	{Kind: "ConfigMap", Name: "istio-ca-root-cert"},
+	{Kind: "ConfigMap", Name: "kube-root-ca.crt"},
+}
+
+var ExclusionByLabels = []ExclusionByLabelsSpec{
+	{Kind: "ServiceAccount", Labels: map[string]string{"cattle.io/creator": "norman"}},
+	{Group: "rbac.authorization.k8s.io", Kind: "RoleBinding", Labels: map[string]string{"cattle.io/creator": "norman"}},
+	{Group: "rbac.authorization.k8s.io", Kind: "Role", Labels: map[string]string{"cattle.io/creator": "norman"}},
+	{Group: "networking.k8s.io", Kind: "NetworkPolicy", Labels: map[string]string{"cattle.io/creator": "norman"}},
```
adrianludwin, line 196: Is there ever a case where we _wouldn't_ want to exclude something with this label? If so, maybe take out `Group` and `Kind` for now, and just test the label? I see you've added a test for this (which is great) but it might be worth adding a comment why only these _specific_ Kinds shouldn't be propagated.

Also, if you could link to a site that defines what `cattle.io/creator:norman` is, that'd be appreciated.

internal/integtest/helpers.go
```diff
@@ -313,6 +313,19 @@ func MakeObjectWithAnnotation(ctx context.Context, resource string, nsName,
 	createdObjects = append(createdObjects, inst)
 }
 
+// MakeObjectWithLabel creates an empty object with label given kind in a specific
+// namespace. The kind and its corresponding GVK should be included in the GVKs map.
+func MakeObjectWithLabel(ctx context.Context, resource string, nsName,
```
adrianludwin, line 318: Should be plural (e.g. `MakeObjectWithLabels`) since it can accept multiple labels. Feel free to fix the function above as well if you feel like it but you don't have to :)

internal/objects/reconciler_test.go
```diff
@@ -411,6 +411,27 @@ var _ = Describe("Basic propagation", func() {
...
 
+	It("should not propagate builtin exclusions by labels", func() {
+		SetParent(ctx, barName, fooName)
+		MakeObjectWithLabel(ctx, "roles", fooName, "role-with-labels-blocked", map[string]string{"cattle.io/creator": "norman"})
+		MakeObjectWithLabel(ctx, "roles", fooName, "role-with-labels-wrong-value", map[string]string{"cattle.io/creator": "testme"})
+		MakeObjectWithLabel(ctx, "roles", fooName, "role-with-labels-something", map[string]string{"app": "hello-world"})
+		AddToHNCConfig(ctx, api.RBACGroup, api.RoleKind, api.Propagate)
+
+		// the first one should not propagate, everything else should
+		Eventually(HasObject(ctx, "roles", barName, "role-with-labels-blocked")).Should(BeFalse())
```
adrianludwin, line 422: This should be `Consistently()`, not `Eventually()`. If HNC's broken and the object _is_ propagated by mistake, this line could happen to run before HNC manages to propagate it, so the test would pass incorrectly. By using `Consistently()`, HNC will keep retrying this for a few seconds (I think we've configured it as 4s), which greatly reduces the chance that the test would mistakenly pass due to a race condition.

Alternatively, you could leave this as `Eventually` but move it down by two lines, after we _know_ that at least two other objects have been propagated. That also makes it far less likely that the test passes by mistake, and will get rid of the extra 4s wait. I'm ok with either.

adrianludwin (commented):
Thanks for this change, it looks great! I've left a few comments the we're good to go. My overall comment is just to not add complexity where we don't need it yet, e.g. if we're _only_ excluding ConfigMaps by name, we don't need to also test their groups/kinds as well. We can easily add that in the future if we need to.

Also, can you please change the title of the commit / PR to explicitly call out Rancher. I have no concerns about hardcoding this for now, we can always move it to a command-line option or even to `HNCConfiguration` later if needed.

/ok-to-test

nobbs:
@adrianludwin Thank you for the thorough review! I've made the changes you requested - you are certainly right, that complexity that is not required should not be added. As this change would be enough to

internal/selectors/selectors.go
```diff
@@ -158,19 +158,43 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
...
+	// exclusion by name
+	for _, excludedResourceName := range cmExclusionsByName {
 		if group == "" && kind == "ConfigMap" && name == excludedResourceName {
 			return true, nil
 		}
 	}
+
+	// exclusion by labels
+	for _, res := range exclusionByLabels {
+		gotLabelValue, ok := labels[res.Key]
```
adrianludwin, line 193: Maybe comment that label values can be empty so we need to check for presence explicitly (I actually thought we could simplify this to `if labels[res.Key] == res.Value`, but I was wrong, and I don't want someone else to "simplify" this in the future by mistake).

adrianludwin (commented):
One last tiny nit (sorry!), plus can you please update the docs at https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#built-in-exceptions to mention this?

internal/selectors/selectors.go
nobbs: Yeah, valid point. done!

nobbs:
> One last tiny nit (sorry!), plus can you please update the docs at https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#built-in-exceptions to mention this?

Done. And there is nothing to be sorry for ;)

adrianludwin (commented):
Thanks for this change!

/lgtm
/approve

PR #127, opened 2022-01-23
Fix timeout stopping testEnv api-server on K8s 1.20+

erikgb:
I am having issues running `make test` locally, and it seems very similar to the issues we had when upgrading our controller dev-environment to K8s 1.21. There is a closed issue in controller-runtime (https://github.com/kubernetes-sigs/controller-runtime/issues/1571) that describes this issue in more detail. In particular this comment: https://github.com/kubernetes-sigs/controller-runtime/issues/1571#issuecomment-945535598.

This PR starts the controller-manager with a cancellable context, and this context is cancelled just before shutting down testEnv.

### DIFF
--- a/internal/integtest/setup.go
+++ b/internal/integtest/setup.go
@@ -16,6 +16,7 @@ limitations under the License.
 package integtest
 
 import (
+	"context"
 	"path/filepath"
 	"testing"
 	"time"
@@ -44,8 +45,9 @@ import (
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
 var (
-	K8sClient client.Client
-	testEnv   *envtest.Environment
+	K8sClient          client.Client
+	testEnv            *envtest.Environment
+	k8sManagerCancelFn context.CancelFunc
 )
 
 func HNCRun(t *testing.T, title string) {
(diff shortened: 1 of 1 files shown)


internal/integtest/setup.go
```diff
@@ -46,6 +47,7 @@ import (
 var (
 	K8sClient client.Client
 	testEnv   *envtest.Environment
+	cancel    context.CancelFunc
```
adrianludwin, line 50: Maybe give this a slightly more descriptive name, e.g. `testEnvCancelFn`? I know that Go prefers short variable names but this seems a bit too terse and made the code a bit hard to understand IMO.

adrianludwin (commented):
Looks great with one minor change, thanks!

/ok-to-test

internal/integtest/setup.go
erikgb: I agree with renaming the variable, but I think `k8sManagerCancelFn` is more correct, isn't it?

adrianludwin:
I didn't think about it too hard so that sgtm too!

internal/integtest/setup.go
erikgb: I have renamed it as suggested, but please let me know if you have changed your mind. 😉

internal/integtest/setup.go
```diff
@@ -108,12 +113,13 @@ func HNCBeforeSuite() {
 	Expect(K8sClient).ToNot(BeNil())
 
 	go func() {
-		err = k8sManager.Start(ctrl.SetupSignalHandler())
```
adrianludwin, line 111: Hmm, sudden thought. Instead of creating the cancelable context from `context.Background()`, should it be from `ctrl.SetupSignalHandler()` instead? What are we losing by not using this?

erikgb: Good idea - if it works. I can see if that works in my environment.

erikgb: PTAL

erikgb: I am uncertain if my latest change (moving the initialization of the cancel-func) is a good idea. Ginkgo seems to want to run AfterSuite not matter what happens. And the cancel-func might be uninitialized (`nil`).

Using `ctrl.SetupSignalHandler()` instead of creating a new context works!

adrianludwin: What about checking whether `k8sManagerCancelFn` is nil or not before calling it in `HNCAfterSuite`? Otherwise this lgtm.

erikgb: Done

Issue #128, opened 2022-01-23
HierarchyConfiguration labels/annotations should be validated on admission


erikgb:
A new feature allowing labels/annotations to be configured in `HierarchyConfiguration` was introduced in https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/122. Both labels and annotations have quite strict validation rules in K8s. To avoid this failing on reconcile of namespaces, we should validate the new fields on admission. I think we should add at least the following validations:

- Valid label key
- Valid label value
- Valid annotation key
- Valid annotation value
- Validate that all labels are managed labels
- Validate that all annotations are managed annotations

Part of #47 

/assign

PR #132, opened 2022-01-24
Improve webhooks deny invalid fields

erikgb:
Started implementing https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/128, I realize that we currently have limited possibilities to report multiple invalid field validation errors in the same admission response. And with the quite detailed validation rules required for managed labels/annotations, I think we need an improvement. At least from an end-user perspective.

This PR suggests to replace https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/bfe42e2ec1c303709373d12d0b3d4d802b92c9c7/internal/webhooks/webhooks.go#L61 with a more generic DenyFromStatus inspired by existing code in controller-runtime (to support implementing high-level admission controllers), using errors/validations from api-machinery. This allows for aggregating multiple webhook validation errors of the same type (defined by statusCode), and should also IMO provide better error messages: Example: `HNCConfiguration.hnc.x-k8s.io \"config\" is invalid: spec.resources[0]: Invalid value: crontabs: not found in the apiserver with error: crontabs does not exist` instead of `Cannot find the crontabs in the apiserver with error: crontabs does not exist`

Also cleaning up some duplicated code I found when searching for for usages of `metav1.StatusReasonInvalid`.

### DIFF
--- a/internal/anchor/validator.go
+++ b/internal/anchor/validator.go
@@ -7,8 +7,11 @@ import (
 
 	"github.com/go-logr/logr"
 	k8sadm "k8s.io/api/admission/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
@@ -81,8 +84,15 @@ func (v *Validator) handle(req *anchorRequest) admission.Response {
 
 	errStrs := validation.IsDNS1123Label(cnm)
 	if len(errStrs) != 0 {
-		msg := fmt.Sprintf("Subnamespace %s is not a valid namespace name: %s", cnm, strings.Join(errStrs, "; "))
-		return webhooks.Deny(metav1.StatusReasonInvalid, msg)
+		fldPath := field.NewPath("metadata", "name")
+		msg := fmt.Sprintf("not a valid namespace name: %s", strings.Join(errStrs, "; "))
+		allErrs := field.ErrorList{
+			field.Invalid(fldPath, cnm, msg),
+		}
+
+		gk := schema.GroupKind{Group: api.GroupVersion.Group, Kind: "SubnamespaceAnchor"}
+		err := apierrors.NewInvalid(gk, cnm, allErrs)
+		return webhooks.DenyFromAPIError(err)
 	}
 
 	switch req.op {
--- a/internal/hierarchyconfig/validator.go
+++ b/internal/hierarchyconfig/validator.go
@@ -15,7 +15,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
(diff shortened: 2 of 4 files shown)


internal/anchor/validator.go
```diff
@@ -81,8 +84,15 @@ func (v *Validator) handle(req *anchorRequest) admission.Response {
...
-		return webhooks.Deny(metav1.StatusReasonInvalid, msg)
+		fldPath := field.NewPath("metadata", "name")
+		msg := fmt.Sprintf("not a valid namespace name: %s", strings.Join(errStrs, "; "))
+		allErrs := field.ErrorList{
+			field.Invalid(fldPath, cnm, msg),
+		}
+
+		gk := schema.GroupKind{Group: api.GroupVersion.Group, Kind: "SubnamespaceAnchor"}
+		err := apierrors.NewInvalid(gk, cnm, allErrs)
+		return webhooks.DenyFromStatusError(err)
```
adrianludwin, line 95: The term "status" (as in, `DenyFromStatusError`) seems a bit weird here. Maybe call this `DenyFromAPIError`? I know "status" is technically what the struct is called, but from the PoV of the client, it's from a package called `.../api/errors` so I think this name fits better.

internal/hierarchyconfig/validator.go
```diff
@@ -525,69 +525,3 @@ func allow(msg string) admission.Response {
 		},
 	}}
 }
-
-// deny is a replacement for controller-runtime's admission.Denied() that allows you to set _both_ a
```
adrianludwin, line 529: Thanks for this cleanup

internal/hncconfig/validator.go
```diff
@@ -96,31 +97,38 @@ func (v *Validator) handle(ctx context.Context, inst *api.HNCConfiguration) admi
...
 		}
 
 		// Validate the type exists in the apiserver. If yes, convert GR to GVK. We
 		// use GVK because we will need to checkForest() later to avoid source
 		// overwriting conflict (forest uses GVK as the key for object reconcilers).
 		gvk, err := v.translator.gvkForGR(gr)
 		if err != nil {
-			return webhooks.DenyInvalid(field,
-				fmt.Sprintf("Cannot find the %s in the apiserver with error: %s", gr, err.Error()))
+			fldErr := field.Invalid(fldPath, gr, fmt.Sprintf("not found in the apiserver with error: %s", err.Error()))
```
adrianludwin, line 115: Better: "unknown resource: %s"

adrianludwin (commented):
This looks like a good change to me so far. I'd ask you to run the full set of e2e tests since those are the only way we can check whether we're getting reasonable error message back. To do this:

* Set your kubecontext to an appropriate cluster
* `make deploy`
* `make test-e2e`

You can ignore the warnings about "repairing HNC," since your code shouldn't affect things like HNC upgrades.

internal/anchor/validator.go
erikgb: Renamed

internal/hncconfig/validator.go
```diff
@@ -96,31 +97,38 @@ func (v *Validator) handle(ctx context.Context, inst *api.HNCConfiguration) admi
...
+	allErrs := field.ErrorList{}
 	for i, r := range inst.Spec.Resources {
 		gr := schema.GroupResource{Group: r.Group, Resource: r.Resource}
-		field := field.NewPath("spec", "resources").Index(i)
+		fldPath := field.NewPath("spec", "resources").Index(i)
 		// Validate the type configured is not an HNC enforced type.
 		if api.IsEnforcedType(r) {
-			return webhooks.DenyInvalid(field, fmt.Sprintf("Invalid configuration of %s in the spec, because it's enforced by HNC "+
-				"with 'Propagate' mode. Please remove it from the spec.", gr))
+			fldErr := field.Invalid(fldPath, gr, "enforced by HNC with 'Propagate' mode. Please remove it from the spec.")
```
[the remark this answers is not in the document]
erikgb: The field value is added to the message by `field.Invalid`, but I modified it according to your suggesting. Example from unit-tests:
````
HNCConfiguration.hnc.x-k8s.io \"config\" is invalid: [spec.resources[0]: Invalid value: roles.rbac.authorization.k8s.io: always uses the 'Propagate' mode and cannot be configured, spec.resources[1]: Invalid value: rolebindings.rbac.authorization.k8s.io: always uses the 'Propagate' mode and cannot be configured]
````

internal/hncconfig/validator.go
erikgb: The field value is added to the message by field.Invalid, but I modified it according to your suggestion. Example from unit-tests:
````
HNCConfiguration.hnc.x-k8s.io \"config\" is invalid: spec.resources[0]: Invalid value: crontabs: unknown resource
````

erikgb:
@adrianludwin I have addressed your initial comments. PTAL!

adrianludwin:
Nice!

/lgtm
/approved
/ok-to-test

PR #133, opened 2022-01-26
Update dependencies for K8s 1.22+

adrianludwin:
The current version of cert-controller won't work on K8s 1.22+ since it
uses the removed v1beta CRD API. This change picks up the latest
cert-controller and all the other key latest dependencies.

Key updates:
* controller-runtime to v0.11
* cert-controller to the latest to pick up CRD v1beta removal
* K8s client libs to 1.23
* controller-tools to latest
* Various other small changes

Tested: smoke tests


erikgb:
FYI: We have experienced issues upgrading our controllers to controller-runtime 0.11.0 and K8s 1.23. See https://github.com/kubernetes-sigs/controller-runtime/issues/1786. Might not be an issue here, but we decided to revert the upgrade - at least for now.

PR #134, opened 2022-01-27
Add validation of managed labels/annotations

### LINKED ISSUE #128: HierarchyConfiguration labels/annotations should be validated on admission

erikgb:
Closes #128. Part of #47.

This adds webhook validation of managed labels and annotations.

Tested: Added unit tests for validations. Ran e2e-tests, but no additions/modifications in this PR.

/cc @adrianludwin

### DIFF
--- a/internal/hierarchyconfig/validator.go
+++ b/internal/hierarchyconfig/validator.go
@@ -12,9 +12,12 @@ import (
 	authzv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -127,6 +130,13 @@ func (v *Validator) handle(ctx context.Context, log logr.Logger, req *request) a
 	if why := config.WhyUnmanaged(req.hc.Spec.Parent); why != "" {
 		reason := fmt.Sprintf("Namespace %q is not managed by HNC (%s) and cannot be set as the parent of another namespace", req.hc.Spec.Parent, why)
 		return webhooks.Deny(metav1.StatusReasonForbidden, reason)
+	}
+
+	allErrs := validateManagedMeta(req.hc)
+	if len(allErrs) > 0 {
+		gk := schema.GroupKind{Group: api.GroupVersion.Group, Kind: "HierarchyConfiguration"}
+		err := apierrors.NewInvalid(gk, req.hc.Name, allErrs)
+		return webhooks.DenyFromAPIError(err)
 	}
 
 	// Do all checks that require holding the in-memory lock. Generate a list of server checks we
@@ -512,6 +522,49 @@ func (r *realClient) IsAdmin(ctx context.Context, ui *authnv1.UserInfo, nnm stri
 
 	// Extract the interesting result
 	return sar.Status.Allowed, err
+}
+
+func validateManagedMeta(hc *api.HierarchyConfiguration) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	fldPath := field.NewPath("spec", "labels")
+	for _, l := range hc.Spec.Labels {
+		if fldErr := validateMetaKey(l.Key, fldPath); fldErr != nil {
+			allErrs = append(allErrs, fldErr)
+		} else if !config.IsManagedLabel(l.Key) { // Only validate managed if key is valid
+			fldErr := field.Invalid(fldPath, l.Key, "not a managed label and cannot be configured")
+			allErrs = append(allErrs, fldErr)
+		}
+		if fldErr := validateLabelValue(l.Key, l.Value, fldPath); fldErr != nil {
+			allErrs = append(allErrs, fldErr)
+		}
+	}
+
+	fldPath = field.NewPath("spec", "annotations")
+	for _, a := range hc.Spec.Annotations {
+		if fldErr := validateMetaKey(a.Key, fldPath); fldErr != nil {
+			allErrs = append(allErrs, fldErr)
+		} else if !config.IsManagedAnnotation(a.Key) { // Only validate managed if key is valid
+			fldErr := field.Invalid(fldPath, a.Key, "not a managed annotation and cannot be configured")
+			allErrs = append(allErrs, fldErr)
+		}
+		// No validation of annotation values; only limitation seems to be the total length (256Kb) of all annotations
+	}
+	return allErrs
+}
+
+func validateMetaKey(k string, path *field.Path) *field.Error {
+	if errs := validation.IsQualifiedName(k); len(errs) != 0 {
+		return field.Invalid(path, k, strings.Join(errs, "; "))
+	}
+	return nil
+}
+
+func validateLabelValue(k, v string, path *field.Path) *field.Error {
+	if errs := validation.IsValidLabelValue(v); len(errs) != 0 {
+		return field.Invalid(path.Key(k), v, strings.Join(errs, "; "))
+	}
+	return nil
 }
 
 // allow is a replacement for controller-runtime's admission.Allowed() that allows you to set the
--- a/internal/hierarchyconfig/validator_test.go
+++ b/internal/hierarchyconfig/validator_test.go
@@ -17,6 +17,69 @@ import (
 	"sigs.k8s.io/hierarchical-namespaces/internal/objects"
 )
 
+func TestManagedMeta(t *testing.T) {
+	f := foresttest.Create("-") // a
+	h := &Validator{Forest: f}
+	l := zap.New()
+	// For this test we accept any label or annotation not starting with 'h',
+	// to allow almost any meta - except the hnc.x-k8s.io labels/annotations,
+	// which cannot be managed anyway. And allows us to use that for testing.
+	if err := config.SetManagedMeta([]string{"[^h].*"}, []string{"[^h].*"}); err != nil {
+		t.Fatal(err)
+	}
+	defer config.SetManagedMeta(nil, nil)
+
+	tests := []struct {
+		name        string
+		labels      []api.MetaKVP
+		annotations []api.MetaKVP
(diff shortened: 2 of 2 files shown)


adrianludwin (commented):
I'll only be able to look at this tomorrow.

Ryan, if you have a few minutes to check this out and like it, feel free to lgtm/approve without me! Otherwise I'll get it tomorrow.

/cc @rjbez17

erikgb:
FYI: I removed some refactoring/cleanup that can be done in a separate PR. Ready for review.

internal/hierarchyconfig/validator_test.go
```diff
@@ -17,6 +17,101 @@ import (
...
+		{name: "unmanaged label", labels: []api.MetaKVP{{Key: "kubernetes.io/metadata.name"}}},
+		{name: "managed annotation", annotations: []api.MetaKVP{{Key: "annot.com/log-index"}}, allowed: true},
+		{name: "unmanaged annotation", annotations: []api.MetaKVP{{Key: "openshift.io/sa.scc.uid-range"}}},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// Setup
+			g := NewWithT(t)
+			// In this test we only allow labels/annotations with the defined prefixes
+			err := config.SetManagedMeta([]string{"label\\.com/.*"}, []string{"annot\\.com/.*"})
```
erikgb, line 41: I wish we had an alternative to regex here. IMO a prefix- or glob-based pattern would be much more user-friendly here. Prefixed labels/annotations are recommended and very common.

internal/hierarchyconfig/validator.go
```diff
@@ -129,6 +132,38 @@ func (v *Validator) handle(ctx context.Context, log logr.Logger, req *request) a
...
+
+	fldPath = field.NewPath("spec", "annotations")
+	for _, a := range req.hc.Spec.Annotations {
+		if err := validateMetaKey(a.Key, fldPath); err != nil {
+			allErrs = append(allErrs, err)
+		} else if !config.IsManagedAnnotation(a.Key) { // Only validate managed if key is valid
+			fldErr := field.Invalid(fldPath, a.Key, "not a managed annotation and cannot be configured")
+			allErrs = append(allErrs, fldErr)
+		}
+		// No validation for annotation values; only limitation seems to be the total length (256Kb) of all annotations
```
erikgb, line 158:
```suggestion
		// No validation of annotation values; only limitation seems to be the total length (256Kb) of all annotations
```

internal/hierarchyconfig/validator.go
```diff
@@ -129,6 +132,38 @@ func (v *Validator) handle(ctx context.Context, log logr.Logger, req *request) a
 		return webhooks.Deny(metav1.StatusReasonForbidden, reason)
 	}
 
+	allErrs := field.ErrorList{}
```
adrianludwin, line 135: This function's getting a bit long, maybe move this all to either a `validateMangedMeta` or two `validateManagedLabels`/`validateManagedAnnotations` functions?

internal/hierarchyconfig/validator.go
```diff
@@ -129,6 +132,38 @@ func (v *Validator) handle(ctx context.Context, log logr.Logger, req *request) a
...
+
+	fldPath := field.NewPath("spec", "labels")
+	for _, l := range req.hc.Spec.Labels {
+		if err := validateMetaKey(l.Key, fldPath); err != nil {
```
adrianludwin, line 139: Instead of `err`, maybe use `fe` for `field.Error` since it's not a standard Go `error`? Not a huge deal, but whenever I see `err` I just assume it's `error`.

internal/hierarchyconfig/validator.go
adrianludwin: Either works for me

internal/hierarchyconfig/validator_test.go
adrianludwin, line 41: Should have a `defer config.SetManagedMeta(nil, nil)` at the beginning of the test to make sure this is properly reset.

Also this is the same for all test cases so I'd set it outside of the loop.

adrianludwin: Discussed on Slack, maybe add an issue for this? Internally we'll almost certainly stick to regexes but we could have a more user-friendly way to set the command lines (just like, for exceptions, we have the generic `/select` annotation but also `/treeSelect`).

internal/hierarchyconfig/validator_test.go
```diff
@@ -17,6 +17,101 @@ import (
...
+
+			got := h.handle(context.Background(), l, req)
+
+			logResult(t, got.AdmissionResponse.Result)
+			g.Expect(got.AdmissionResponse.Allowed).Should(Equal(tc.allowed))
+		})
+	}
+}
+
+func TestManagedMetaSyntax(t *testing.T) {
```
adrianludwin, line 59: These two tests (...Permissions and ...Syntax) are basically identical. I'd just combine them myself. The structure of the test is basically the same and they're testing the same function.

adrianludwin (commented):
I think this is sufficient testing. We generally don't have e2e tests (which is what I think you mean by "integration tests"?) for webhook validation except to make sure that it's set up at all. Unit tests are more than enough to test whether the logic is correct.

Otherwise, this lgtm apart from some small requests. Thanks!

internal/hierarchyconfig/validator_test.go
erikgb: Created issue.

internal/hierarchyconfig/validator.go
erikgb: I went for `fldErr`, and hope that is ok. Two letter variable names can be a bit "tricky". PTAL!

adrianludwin: np!

internal/hierarchyconfig/validator.go
```diff
@@ -129,6 +132,13 @@ func (v *Validator) handle(ctx context.Context, log logr.Logger, req *request) a
 		return webhooks.Deny(metav1.StatusReasonForbidden, reason)
 	}
 
+	allErrs := validateManagedMeta(req.hc)
+	if len(allErrs) > 0 {
+		gk := schema.GroupKind{Group: api.GroupVersion.Group, Kind: "HierarchyConfiguration"}
+		err := apierrors.NewInvalid(gk, req.hc.Name, allErrs)
+		return webhooks.DenyFromAPIError(err)
+	}
+
```
adrianludwin, line 141: Can we shorten this even more? I like the lower-level function to encapsulate as much information as possible so that the top-level function is easier to read, e.g:

```
if denied := validateManagedMeta(req.hc); denied != nil {
  return denied
}
```

Now the caller (`handle()`) doesn't have to know anything about the _type_ of information coming out of `validateManagedMeta` other than that it's ready to be returned to _its_ caller (`Handle()`).

internal/hierarchyconfig/validator_test.go
```diff
@@ -17,6 +17,71 @@ import (
...
+func TestManagedMeta(t *testing.T) {
+	f := foresttest.Create("-") // a
+	h := &Validator{Forest: f}
+	l := zap.New()
+	// For this test we accept any label or annotation not starting with 'h',
+	// to allow almost any meta - except the hnc.x-k8s.io labels/annotations,
+	// which cannot be managed anyway. And allows us to use that for testing.
+	if err := config.SetManagedMeta([]string{"[^h].*"}, []string{"[^h].*"}); err != nil {
+		panic(err)
+	}
```
adrianludwin, line 29: Why `panic`? I think `t.Fatal(err)` is probably better?

internal/hierarchyconfig/validator_test.go
```diff
@@ -17,6 +17,71 @@ import (
...
+	defer func() {
+		_ = config.SetManagedMeta(nil, nil)
+	}()
```
adrianludwin, line 32: Why not `defer config.SetManagedMeta(nil, nil)`? Is the anonymous function adding anything?

internal/hierarchyconfig/validator_test.go
```diff
@@ -17,6 +17,71 @@ import (
...
+	tests := []struct {
+		name        string
+		labels      []api.MetaKVP
+		annotations []api.MetaKVP
+		allowed     bool
+	}{
+		{name: "managed label", labels: []api.MetaKVP{{Key: "label.com/team"}}, allowed: true},
+		{name: "unmanaged label", labels: []api.MetaKVP{{Key: api.LabelIncludedNamespace}}},
+		{name: "managed annotation", annotations: []api.MetaKVP{{Key: "annot.com/log-index"}}, allowed: true},
+		{name: "unmanaged annotation", annotations: []api.MetaKVP{{Key: api.AnnotationManagedBy}}},
```
adrianludwin, line 43: Nit: use the "ok:" and "invalid" prefix in the names for consistency with the other cases in this test.

adrianludwin (commented):
Sorry four last nits! LMK if you disagree with them.

internal/hierarchyconfig/validator.go
erikgb: There reason I left the `allErrs` there, is to make it more obvious how to include more field errors in the same response. And to avoid more refactoring if/when more fields are added to `HierarchyConfiguration` that require validation in this webhook.

If we were using the high-level webhook framework provided by controller-runtime, `Error`/`field.ErrorList` is the "contract" between your webhook implementation and controller-runtime. That is another reason for keeping it as suggested IMO.

So here I do not agree with you, and would like to keep it as is. But if you insist..... 😉

internal/hierarchyconfig/validator_test.go
erikgb: Thanks for learning me how to do this! We use mostly Gomega/Ginkgo, and I have only done Golang for a year or so....

internal/hierarchyconfig/validator_test.go
erikgb: Just my IDE complaining about unhandled error, but in this case it's "wrong". 😅

internal/hierarchyconfig/validator.go
adrianludwin: Interesting. Let's leave it as-is for now, but generally I prefer to keep the current state of the code as readable as possible, and save the flexibility until we actually use them in the PRs. But especially if you'll be adding that shortly, it's fine to go in like this.

adrianludwin (commented):
/lgtm
/approve

PR #136, opened 2022-01-29
Use DynamicRESTMapper to lookup metadata from apiserver

erikgb:
This is mainly a refactoring of existing code, but modifies the internal machinery slightly. The initial motivation was to add more apiserver lookups to improve webhook validations in the `objects` package. Since the existing code performing similar stuff was private to the `hncconfig` package I had to do some changes anyway.

It turns out that controller-runtime provides a `DynamicRESTMapper`, that I think will fit very well in HNC for looking up resource information from the apiserver. HNC needs dynamic information in runtime, and that is exactly what this class provides. It also has some additional configuration features that can eventually be used in a follow-up PR.

Testing: Ran all tests locally with success, but I do not currently have a cluster available to test this change more. I expect it to work well, but doing changes to the machinery is always a bit risky.

/cc @adrianludwin

### DIFF
--- a/internal/apimeta/mapper.go
+++ b/internal/apimeta/mapper.go
@@ -1,0 +1,74 @@
+package apimeta
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// NotNamespacedError is returned if the mapper can find the resource, but it is not namespaced.
+type NotNamespacedError struct {
+	GroupResource schema.GroupResource
+}
+
+func (e *NotNamespacedError) Error() string {
+	return fmt.Sprintf("resource %q is not namespaced", e.GroupResource)
+}
+
+func IsNotNamespacedError(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err.(type) {
+	case *NotNamespacedError:
+		return true
+	default:
+		return false
+	}
+}
+
+func NewGroupKindMapper(c *rest.Config) (*GroupKindMapper, error) {
+	restMapper, err := apiutil.NewDynamicRESTMapper(c)
+	if err != nil {
+		return nil, err
+	}
+	return &GroupKindMapper{
+		RestMapper: restMapper,
+	}, nil
+
+}
+
+// GroupKindMapper allows callers to map resources to kinds using apiserver metadata at runtime
+type GroupKindMapper struct {
+	RestMapper meta.RESTMapper
+}
+
+// NamespacedKindFor maps namespaced GR to GVK by using a DynamicRESTMapper
+// to discover resource types at runtime. Will return an error if the resource isn't namespaced.
+func (r *GroupKindMapper) NamespacedKindFor(gr schema.GroupResource) (schema.GroupVersionKind, error) {
+	resource := gr.WithVersion("")
+	kinds, err := r.RestMapper.KindsFor(resource)
+	if err != nil {
+		return schema.GroupVersionKind{}, err
+	}
+	// Not sure if this can ever happen? RestMapper.KindsFor does not specify if no
+	// matches will error out or not, so just making sure we do not get issues.
+	if len(kinds) == 0 {
+		return schema.GroupVersionKind{}, &meta.NoResourceMatchError{PartialResource: resource}
+	}
+
+	// KindsFor returns the list of potential kinds in priority order
+	gvk := kinds[0]
+	mapping, err := r.RestMapper.RESTMapping(gvk.GroupKind())
+	if err != nil {
+		return schema.GroupVersionKind{}, fmt.Errorf("while attempting to determine if kind is namespaced: %w", err)
+	}
+	if mapping.Scope.Name() != meta.RESTScopeNameNamespace {
+		return schema.GroupVersionKind{}, &NotNamespacedError{GroupResource: gr}
+	}
+
+	return gvk, nil
+}
--- a/internal/hncconfig/reconciler.go
+++ b/internal/hncconfig/reconciler.go
@@ -8,23 +8,21 @@ import (
 	"sync"
 	"time"
 
-	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/restmapper"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
-
 	"github.com/go-logr/logr"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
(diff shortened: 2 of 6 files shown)


internal/apimeta/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// ResourceMapper allows callers to map resources to kinds.
+type ResourceMapper interface {
+	// NamespacedKindFor returns the priority mapping GVK for a namespaced GR.
+	// If the GR doesn't exist, or isn't namespaced, returns an error.
+	NamespacedKindFor(gr schema.GroupResource) (schema.GroupVersionKind, error)
+}
+
```
adrianludwin, line 18: Generally speaking, the producer module shouldn't declare the interface type: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

Right now, only `internal/hncconfig` uses this type so I think it's fine to move this interface to that package (and make it private). Maybe we can move it back here if a bunch of other modules start implementing it to avoid having to redefine it multiple times.

internal/apimeta/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	switch err.(type) {
+	case *NotNamespacedError:
+		return true
+	default:
+		return false
+	}
+}
+
+func NewAPIMapper(c *rest.Config) (ResourceMapper, error) {
```
adrianludwin, line 40: Bikeshedding time - instead of `APIMapper` (which is very generic), how about `GroupKindMapper`? That's far more descriptive of what it actually does.

You can then call the variables or fields `gkMapper` which might read a little better than simply `mapper`.

internal/apimeta/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// APIMapper allows callers to discover apiserver metadata at runtime
+type APIMapper struct {
+	RestMapper meta.RESTMapper
+}
+
+// NamespacedKindFor implements ResourceMapper.NamespacedKindFor by using a DynamicRESTMapper
+// to discover resource types at runtime.
+func (r *APIMapper) NamespacedKindFor(gr schema.GroupResource) (schema.GroupVersionKind, error) {
```
adrianludwin, line 58: I don't really see the need to have `Namespaced` in the name? You can just comment it. Also this doesn't _just_ return the kind, so maybe `GetGVK` or (if you think it's important to keep) `GetNamespacedGVK`? (GVK is a fairly well-known acronym for GroupVersionKind).

internal/apimeta/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+	// matches will error out or not, so just making sure we do not get issues.
+	if len(kinds) == 0 {
+		return schema.GroupVersionKind{}, &meta.NoResourceMatchError{PartialResource: resource}
+	}
+
+	// KindsFor returns the list of potential kinds in priority order
+	gvk := kinds[0]
+	mapping, err := r.RestMapper.RESTMapping(gvk.GroupKind())
+	if err != nil {
+		return schema.GroupVersionKind{}, err
```
adrianludwin, line 74: Maybe wrap the error, e.g. `"while attempting to determine if kind is namespaced: %w"`

internal/hncconfig/reconciler.go
```diff
@@ -145,38 +137,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		// Look if the resource exists in the API server.
-		gvk, err := gvkForGR(gr, allRes)
+		gvk, err := r.ResourceMapper.NamespacedKindFor(gr)
 		if err != nil {
-			// If the type is not found, log error and write conditions and return the
+			// Error looking up enforced types. Log error, write conditions and return the
 			// error for a retry.
 			r.Log.Error(err, "while trying to reconcile the enforced resource", "resource", gr)
+			// TODO(adrianludwin): Should we rather use api.ReasonUnknown here?
 			r.writeCondition(inst, api.ConditionBadTypeConfiguration, api.ReasonResourceNotFound, err.Error())
```
adrianludwin, line 166: Yeah probably a good idea, you can go ahead and change that. Or maybe factor out the code at line 203 and use it here as well. Thanks!

internal/apimeta/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+	NamespacedKindFor(gr schema.GroupResource) (schema.GroupVersionKind, error)
+}
+
+// NotNamespacedError is returned if the mapper can find the resource, but it is not namespaced.
+type NotNamespacedError struct {
+	GroupResource schema.GroupResource
+}
+
+func (e *NotNamespacedError) Error() string {
+	return fmt.Sprintf("resource %v is not namespaced", e.GroupResource)
```
adrianludwin, line 25: I think I'd prefer `%q` here to `%v` to maintain the double quotes. We switched a bunch of places to `%q` throughout HNC as we found it made the errors easier to read.

internal/hncconfig/reconciler_test.go
```diff
@@ -301,7 +301,7 @@ var _ = Describe("HNCConfiguration", func() {
 		AddToHNCConfig(ctx, api.RBACGroup, "clusterroles", api.Propagate)
 
 		Eventually(GetHNCConfigCondition(ctx, api.ConditionBadTypeConfiguration, api.ReasonResourceNotNamespaced)).
-			Should(ContainSubstring("Resource \"clusterroles.rbac.authorization.k8s.io\" is not namespaced"))
+			Should(ContainSubstring("resource clusterroles.rbac.authorization.k8s.io is not namespaced"))
```
adrianludwin, line 304: If you use `%q` above we'll keep the double quotes. +1 to correcting the capitalization though, thanks.

adrianludwin (commented):
Ok I was a bit nervous about this and I'd _ideally_ like to see it tested on a real cluster, but I suppose our postsubmits run it on Kind so it's not the end of the world if we submit it before testing it too much. I can try it out on GKE later as well.

/cc @rjbez17 
Ryan, any concerns?

lgtm after a few changes
/ok-to-test

internal/apimeta/mapper.go
erikgb: The term `Namespaced` was introduced to make the caller aware that the func will error out if you ask for a non-namespaced GVK-mapping.

I am personally not a big fan of the `Get`/`Set` func-names - unless they are **only**  getting/setting. So I tried to build on the IMO meaningful function names in `RestMapper`.

internal/apimeta/mapper.go
erikgb: One of the reasons I created this PR in the first place, was exactly to map more that just GR to (namespaced) GVK. 🙄 And to factor out some low-level technical code from the reconcilers - to make it easier to read/understand. But I can probably try to improve the validation of objects without doing anything with this mapping....

adrianludwin: Ok, up to you. I'd be happy to change the name to something specific today and then generalize it when it does more (these are all fully internal names, after all) but I don't feel too strongly.

internal/apimeta/mapper.go
adrianludwin: I'd kind of like this function to only return the `Kind` if it keeps the current name, or include `GVK` in its name if it continues to return a GVK. Beyond that, up to you.

internal/apimeta/mapper.go
erikgb: Thx, I wasn't aware of this practice in Golang. Fixed.

internal/apimeta/mapper.go
erikgb: Changed name to `GroupKindMapper` - as suggested.

internal/apimeta/mapper.go
erikgb: It returns a [kind](https://kubernetes.io/docs/reference/using-api/api-concepts/#standard-api-terminology), fully qualified as a GVK. 😉  I do not see the need to use the GVK term when naming a function. The return type is quite clear. I prefer including "namespaced" in the name, but I'll change this func-name to whatever you wish - to get this merged. 😃

internal/apimeta/mapper.go
erikgb: Done, or were you thinking of another form of wrapping. Please let know if I didn't do it correctly.

internal/hncconfig/reconciler.go
erikgb: Good suggesting to align this with the other use of the function. I'll have to think a bit how to do it...

internal/apimeta/mapper.go
erikgb: Thx, fixed!

adrianludwin (commented):
lgtm once you decide what to do about that last refactoring, thanks!

internal/apimeta/mapper.go
adrianludwin: lgtm

internal/hncconfig/reconciler.go
erikgb: I have added a suggestion. Please let me know what you think!

adrianludwin (commented):
/lgtm
/approve

Thanks for this change!

PR #137, opened 2022-01-30
Refactor webhook deny responses (cont)

erikgb:
The goal of this PR is to eliminate callers to the recently deprecated generic `webhooks.Deny` internal function, and remove the function. I will prepare follow-up PR(s), where the ultimate goals is to provide the end-user with validation errors that:

- aggregates multiple validation errors if appropriate (and not the current fail-first behavior)
- are better structured and follows an established pattern in HNC
- are generally better aligned with standard Kubernetes validation errors

This is a continuation of https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/132, that removes almost all callers to the generic `webhooks.Deny` func that got deprecated in this PR. Still some callers left in the "objects" validator, that I will fix in a follow-up PR after some required refactoring. This change is mainly refactoring existing code, but some error messages might change slightly.

### DIFF
--- a/internal/webhooks/webhooks.go
+++ b/internal/webhooks/webhooks.go
@@ -8,6 +8,8 @@ import (
 	authnv1 "k8s.io/api/authentication/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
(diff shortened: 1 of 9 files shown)


internal/webhooks/webhooks.go
```diff
@@ -43,16 +45,44 @@ func Allow(msg string) admission.Response {
 	}}
 }
 
+func DenyBadRequest(err error) admission.Response {
+	return denyFromAPIError(apierrors.NewBadRequest(err.Error()))
+}
+
+func DenyConflict(qualifiedResource schema.GroupResource, name string, err error) admission.Response {
```
adrianludwin, line 52: I'm wondering if `name` is underspecified for most of the callers of this function (ditto for `DenyForbidden`)? E.g. an anchor name doesn't say what parent namespace it's in, and this call:

`return webhooks.DenyForbidden(api.HierarchyConfigurationGR, api.Singleton, err)`

doesn't indicate the namespace at all. Perhaps we should update the callers to either include the namespace name or accept two parameters to this function and create a string like "namespace-name/object-name" if both are provided?

Or maybe it doesn't matter because the namespace names are already in the error messages? wdyt?

adrianludwin (commented):
lgtm after one concern

internal/webhooks/webhooks.go
erikgb: I understand your concern here, and for sure the error messages and internal API for creating the errors can be improved. We could use [NamespacedName](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apimachinery/pkg/types/namespacedname.go#L27) for namespaced resources, but for cluster resources that does not make sense. Currently these new API error factory methods are just delegating into the corresponding function in api-machinery.

I personally try to avoid having multiple parameters of the same type in functions, mainly because I think positional parameters are error-prone. But that is a personal preference. 😅 

Since this PR is a refactor PR, I would prefer leaving it as is for now. I never had the ambition to make all error messages "perfect" here, but I am quite sure I am not making the error messages worse in this PR!

adrianludwin: Fair enough, thanks.

adrianludwin (commented):
/lgtm
/approve

PR #138, opened 2022-01-30
Eliminate some warnings in Go code

erikgb:
After a couple of PRs, I notice that my IDE (Goland) "complains" about quite a few things in the existing HNC codebase.

This PR fixes some (hopefully) non-controversial warnings emitted after analyzing the code.

Testing: Tests run successfully on my local workstation.

/kind cleanup


internal/forest/namespacestructure_test.go
```diff
@@ -74,10 +74,10 @@ func TestDescendantNames(t *testing.T) {
...
 	//         789
 	kinship := map[string][]string{
-		"a":   []string{"b", "c"},
-		"b":   []string{"x", "y", "z"},
-		"c":   []string{"123", "456"},
-		"123": []string{"789"},
+		"a":   {"b", "c"},
+		"b":   {"x", "y", "z"},
+		"c":   {"123", "456"},
+		"123": {"789"},
```
adrianludwin, line 80: Cool! I wonder if this used to be required, I've always hated that boilerplate.

internal/hierarchyconfig/reconciler_test.go
```diff
@@ -32,7 +32,7 @@ var _ = Describe("Hierarchy", func() {
 		fooName = CreateNS(ctx, "foo")
 		barName = CreateNS(ctx, "bar")
 		config.SetNamespaces("")
-		config.SetManagedMeta(nil, nil)
+		Ω(config.SetManagedMeta(nil, nil)).Should(Succeed())
```
adrianludwin, line 35: As cute as it is, we don't use `Ω` anywhere else in our codebase. Can we just use `Expect()` here instead?

internal/kubectl/root.go
```diff
@@ -54,7 +55,7 @@ type Client interface {
 }
 
 func init() {
-	api.AddToScheme(scheme.Scheme)
+	utilruntime.Must(api.AddToScheme(scheme.Scheme))
```
adrianludwin, line 58: Can you explain this one a bit more?

internal/objects/reconciler_test.go
```diff
@@ -215,11 +215,12 @@ var _ = Describe("Exceptions", func() {
 				}
 
 				// update the role with the selector and treeSelector annotation
-				UpdateObjectWithAnnotations(ctx, api.RoleResource, names[p], "testrole", map[string]string{
+				err := UpdateObjectWithAnnotations(ctx, api.RoleResource, names[p], "testrole", map[string]string{
```
adrianludwin, line 218: Are there any cases today where `UpdateObjectWithAnnotations` is _supposed_ to fail? If not, maybe put the `Expect(...).Should(Succeed())` right inside the function?

internal/selectors/selectors.go
```diff
@@ -179,6 +179,7 @@ func isExcluded(inst *unstructured.Unstructured) (bool, error) {
 	name := inst.GetName()
 	kind := inst.GetKind()
 	group := inst.GroupVersionKind().Group
+	// TODO(erikgb): This variable name collides with imported package name
 	labels := inst.GetLabels()
```
adrianludwin, line 183: `instLabels`? Though honestly I think this is only used once (line 194) so there's no need to make a local var for it.

internal/setup/reconcilers.go
```diff
@@ -17,7 +17,10 @@ import (
 //
 // This function is called both from main.go as well as from the integ tests.
 func CreateReconcilers(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, useFakeClient bool) error {
-	crd.Setup(mgr, useFakeClient)
+	err := crd.Setup(mgr, useFakeClient)
+	if err != nil {
```
adrianludwin, line 21: Why not a one-liner?

adrianludwin (commented):
Huh, I could have sworn we had warning turned on at some time in the past but I guess I should have become suspicious because I hadn't seen any warnings in ages.

Is there a way to turn on error-on-warning in the makefile?

Overall looks great (thanks for doing this), just a few questions.

/ok-to-test

internal/kubectl/root.go
erikgb: Just to ensure panic if an error occurs here. The function returns an error, and that gives a warning in my IDE if not handled (or ignored).

erikgb:
@adrianludwin I have addressed all your comments. PTAL!

internal/kubectl/root.go
adrianludwin: Hmm, I might have just handled the error rather than pull in another dependency, I feel like that would be more obvious. LMKWYT and then I'll approve either way.

adrianludwin:
@erikgb Is there a way to turn on warnings in the makefile as well so we don't have these problems creep up again?

internal/kubectl/root.go
erikgb: The dependency is not "new". It is part of api-machinery. :wink But I have no strong feeling about this. Can also just ignore the error unintentionally - as it used to be.

adrianludwin:
> > @erikgb Is there a way to turn on warnings in the makefile as well so we don't have these problems creep up again?
> 
> Probably yes, but I am neither an expert on Makefile nor Go.

np, just filed #146

internal/kubectl/root.go
adrianludwin: I mean there are lots of api-machinery packages we don't import. It's new to this file and that's added complexity. But it's nbd so I'll just approve.

Issue #140, opened 2022-02-01
Could not update the hierarchical configuration - on Openshift 


arturobrzut:
Environment Openshift 4.8
HNC v0.9.0

Hi 
I install HNC on Openshift cluster 4.8
And I have such issue:
```
kubectl create namespace my-parent
namespace/my-parent created

kubectl hns create -n my-parent my-child1   
Successfully created "my-child1" subnamespace anchor in "my-parent" namespace

kubectl create namespace my-child2
namespace/my-child2 created

kubectl hns set my-child2 --parent my-parent
Setting the parent of my-child2 to my-parent

Could not update the hierarchical configuration of my-child2.
Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "my-child2": system:image-pullers (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
  * Namespace "my-child2": system:image-builders (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
  * Namespace "my-child2": system:deployers (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
To fix this, please rename or remove the conflicting objects first.
```

Can we turn off checking some OCP resource in HNC?

adrianludwin:
Are you able to add annotations to the RBAC role bindings? If so, you can
use exceptions
<https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#use-limit-propagation>
to disable annotations for them. Otherwise, is there a systematic way to
identify which objects *shouldn't* be propagated, like a name or a label?
If so, we could add all the necessary OpenShift information here
<https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/082779afa4b5bdd6c9231906f3639f41d79a6999/internal/selectors/selectors.go#L161>
for
the next version of HNC.

arturobrzut:
@adrianludwin 
It works when I add annotations in parent namespace resources

but I want to add PR which exclude this 3 RoleBinding 
I will prepare it

adrianludwin:
What's the criteria you'll use to exclude these? Is it a well-known label?
If so, please include documentation here:
https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#built-in-exceptions

Thanks!

arturobrzut:
[...]

PR #141, opened 2022-02-01
Support for Openshift

arturobrzut:
Support for Openshift
https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/140

Environment Openshift 4.8
HNC v0.9.0

Hi
I install HNC on Openshift cluster 4.8
And I have such issue:
```
kubectl create namespace my-parent
namespace/my-parent created

kubectl hns create -n my-parent my-child1   
Successfully created "my-child1" subnamespace anchor in "my-parent" namespace

kubectl create namespace my-child2
namespace/my-child2 created

kubectl hns set my-child2 --parent my-parent
Setting the parent of my-child2 to my-parent

Could not update the hierarchical configuration of my-child2.
Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "my-child2": system:image-pullers (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
  * Namespace "my-child2": system:image-builders (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
  * Namespace "my-child2": system:deployers (rbac.authorization.k8s.io/v1, Kind=RoleBinding)
To fix this, please rename or remove the conflicting objects first.
```

### DIFF
--- a/internal/selectors/selectors.go
+++ b/internal/selectors/selectors.go
@@ -174,6 +174,18 @@ var exclusionByLabels = []ExclusionByLabelsSpec{
 	{Key: "cattle.io/creator", Value: "norman"},
 }
 
+// A annotation as a key, value pair, used to exclude resources matching this annotation
+type ExclusionByAnnotationsSpec struct {
+	Key   string
+	Value string
+}
+
+// ExclusionByAnnotationsSpec are known annotation key which are excluded from propagation. Right
+// now only used to exclude resources created by Openshift
+var exclusionByAnnotations = []ExclusionByAnnotationsSpec{
+	{Key: "openshift.io/description"},
+}
+
 // isExcluded returns true to indicate that this object is excluded from being propagated
 func isExcluded(inst *unstructured.Unstructured) (bool, error) {
 	name := inst.GetName()
@@ -193,6 +205,16 @@ func isExcluded(inst *unstructured.Unstructured) (bool, error) {
 		// nonexisting key in the `labels` map will also return an empty string ("") - potentially
 		// causing false matches if `ok` is not checked
 		if ok && gotLabelValue == res.Value {
+			return true, nil
+		}
+	}
+
+	// exclusion by annotations
+	for _, res := range exclusionByAnnotations {
+		gotAnnotationValue, ok := inst.GetAnnotations()[res.Key]
+		// we check also if res.Value is an empty string (""),
+		// this is for excluding resources that contain defined keys.
+		if ok && (gotAnnotationValue == res.Value || res.Value == "") {
 			return true, nil
 		}
 	}


internal/selectors/selectors.go
```diff
@@ -161,6 +161,9 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
 // cmExclusionsByName are known (istio and kube-root) CA configmap which are excluded from propagation
 var cmExclusionsByName = []string{"istio-ca-root-cert", "kube-root-ca.crt"}
 
+// cmExclusionsByName are known (istio and kube-root) CA configmap which are excluded from propagation
```
adrianludwin, line 164: Fix comment

adrianludwin (commented):
Are we sure this is the best way to do this? Does Openshift not put labels on these objects?

internal/selectors/selectors.go
arturobrzut: no , openshift does not set labels on it
here is my output
```
oc describe rolebinding -n abc       
Name:         postgresql-operator.v0.1.1
Labels:       <none>
Annotations:  <none>
Role:
  Kind:  Role
  Name:  postgresql-operator.v0.1.1
Subjects:
  Kind            Name                 Namespace
  ----            ----                 ---------
  ServiceAccount  postgresql-operator  


Name:         system:deployers
Labels:       <none>
Annotations:  openshift.io/description:
                Allows deploymentconfigs in this namespace to rollout pods in this namespace.  It is auto-managed by a controller; remove subjects to disa...
              propagate.hnc.x-k8s.io/none: true
Role:
  Kind:  ClusterRole
  Name:  system:deployer
Subjects:
  Kind            Name      Namespace
  ----            ----      ---------
  ServiceAccount  deployer  abc


Name:         system:image-builders
Labels:       <none>
Annotations:  openshift.io/description:
                Allows builds in this namespace to push images to this namespace.  It is auto-managed by a controller; remove subjects to disable.
              propagate.hnc.x-k8s.io/none: true
Role:
  Kind:  ClusterRole
  Name:  system:image-builder
Subjects:
  Kind            Name     Namespace
  ----            ----     ---------
  ServiceAccount  builder  abc


Name:         system:image-pullers
Labels:       <none>
Annotations:  openshift.io/description:
                Allows all pods in this namespace to pull images from this namespace.  It is auto-managed by a controller; remove subjects to disable.
              propagate.hnc.x-k8s.io/none: true
Role:
  Kind:  ClusterRole
  Name:  system:image-puller
Subjects:
  Kind   Name                        Namespace
  ----   ----                        ---------
  Group  system:serviceaccounts:abc  
```

adrianludwin:
How about checking for the presence of an annotation called `
openshift.io/description`? That seems far safer to me IMO since it's
targeted to exactly one vendor.

arturobrzut:
It means 
not propagate all RoleBidning which contains:
  Annotations:  openshift.io/description:
?
I will recheck it

adrianludwin:
I would go a bit further, and say: don't propagate _any_ objects with that
annotation.

arturobrzut:
ok, it looks reasonable.
Should I update this  PR? or it will be done by you?

adrianludwin:
Can you please update the PR? I have no ability to test on Openshift.

arturobrzut:
ok, i will do it tomorrow

arturobrzut:
First proposal
exclude all resources with openshift annotation, no matter what is inside this annotation

```
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ oc create namespace my-parent
namespace/my-parent created
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ oc create namespace my-child 
namespace/my-child created
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ kubectl hns set my-child --parent my-parent
Setting the parent of my-child to my-parent
Succesfully updated 1 property of the hierarchical configuration of my-child
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ oc create namespace my-child
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ kubectl hns create -n my-parent second-child
Successfully created "second-child" subnamespace anchor in "my-parent" namespace
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ oc hns tree  parent               
Error reading namespace parent: namespaces "parent" not found
╭─   ~/dev/g/s/github.c/hierarchical-namespaces on   openshift-support !1 
╰─❯ oc hns tree  my-parent
my-parent
├── my-child
└── [s] second-child

[s] indicates subnamespaces
╭─   ~/dev/go/src/github.com/hierarchical-namespaces on   openshift-support !1 ························································· at  10:44:53
╰─❯ ```

**Comment by adrianludwin** [comment]
/ok-to-test

Can you please squash these two commits so there's only one in the PR? Also, can you please add a note about how you tested this change, i.e. to confirm it actually works on Openshift? Finally, please add any other interesting descriptions to the CL message. A good example is here: https://github.com/kubernetes-sigs/hierarchical-namespaces/commit/485aef19c929d547503ada473c32489f13d0b5ba

@erikgb , I forgot where we left off on Friday. Are you ok with this solution for OpenShift or did you think it was critical to add the name-based resolution instead? I prefer this version for now, at least until we're able to switch a fully command-line option interface.

**Comment by adrianludwin** [comment]
/easycla

Trying to figure out how to change from the old to new CLA format

**Comment by adrianludwin** [comment]
... the presubmits failed, looks like some files are formatted incorrectly. You can use `make fmt` to fix this (do _not_ use `gofmt` as it will modify everything in `/vendor`).

**Comment by arturobrzut** [comment]
I will change it and I also work to add me to EasyCLA as corporate user

**Comment by adrianludwin** [comment]
Out of interest @arturobrzut - we're considering making an (unrelated) change to HNC that would require K8s 1.20 or later. Is that ok for you, or are you on an older version?

**Comment by arturobrzut** [comment]
it is ok

**Comment by adrianludwin** [comment]
I'm ready to merge this; I discussed it with @erikgb and while we agree that we need to stop hardcoding support for K8s distros like OpenShift, we don't want to hold up this change either.

There are only two things left:
* Can you please get the CLA authorized? (I'm guessing you're working on this? There are a bunch of bot comments at the top of this PR that might help)
* Can you please add information on how you tested this change to the commit message (not the PR message)? See https://github.com/kubernetes-sigs/hierarchical-namespaces/commit/485aef19c929d547503ada473c32489f13d0b5ba as an example.

Thanks for this change, I'm excited to see it go in!

**Comment by arturobrzut** [comment]
ok
yes i work on Easy CLA
and i will add comment for commit

**Comment by adrianludwin** [comment]
Thanks!

File: internal/selectors/selectors.go [hunk 2]
```diff
@@ -174,6 +174,18 @@ var exclusionByLabels = []ExclusionByLabelsSpec{
...
+// A annotation as a key, value pair, used to exclude resources matching this annotation
+type ExclusionByAnnotationsSpec struct {
+	Key   string
+	Value string
+}
+
+// ExclusionByAnnotationsSpec are known annotation key which are excluded from propagation. Right
+// now only used to exclude resources created by Openshift
+var exclusionByAnnotations = []ExclusionByAnnotationsSpec{
+	{Key: "openshift.io/description", Value: ""},
```
erikgb, line 186:
```suggestion
	{Key: "openshift.io/description"},
```

internal/selectors/selectors.go
```diff
@@ -197,5 +209,15 @@ func isExcluded(inst *unstructured.Unstructured) (bool, error) {
 		}
 	}
 
+	// exclusion by annotations
+	for _, res := range exclusionByAnnotations {
+		gotAnnotationValue, ok := inst.GetLabels()[res.Key]
+		// we check also if res.Value is an empty string (""),
+		// this is for excluding resources that contain defined keys.
+		if ok && ((gotAnnotationValue == res.Value) || (res.Value == "")) {
```
erikgb, line 217:
```suggestion
		if ok && (gotAnnotationValue == res.Value || res.Value == "") {
```

erikgb (changes requested):
Some minor suggestions, but we need some tests for this new feature and some docs. Maybe use the related PR as inspiration? https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/125/files

internal/selectors/selectors.go
adrianludwin: +1

internal/selectors/selectors.go
adrianludwin: +1

adrianludwin:
> Some minor suggestions, but we need some tests for this new feature and some docs. Maybe use the related PR as inspiration? https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/125/files

Both good points. I guess I was getting a bit lax, thanks Erik!

internal/selectors/selectors.go
arturobrzut: Ok ,
I thought about it and you are probably right. Better to give only the name, than to have ambiguity when giving an empty annotation verse of any annotation
 I will make changes

erikgb:
@arturobrzut You resolved both my code suggestions, but I cannot see a change in the code. Can you fix this, so we can get this merged? LGTM after the small adjustments.

arturobrzut:
@erikgb done

erikgb:
Code LGMT, but I think we should have some tests and docs for this.

adrianludwin:
Actually, I can see from [this comment](https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/141#issuecomment-1027761051) that @arturobrzut did test this and that it's working, so I'm happy to put this in as soon as #160 goes in first.

PR #149, opened 2022-02-24
Copy labels and annotations from SubnamespaceAnchor to child namespace

mishamo:


adrianludwin (commented):
Thanks for this change! Can you please explain a bit about why you want this feature?

But this isn't the UX I'd like to support this feature. The _anchor_ should be able to have labels and annotations on it (e.g. "created-by: some-tool") that are _not_ copied to the child namespaces. This is how labels and annotations work elsewhere in K8s - for example, the labels on a Deployment are _not_ copied to its pods; rather, the podspec in the Deployment includes the labels that should be created on the pod.

In addition, HNC recently added a new feature (not yet released) called _managed labels_ ([link](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#use-managed-labels)) and this feature needs to integrate with that. For example, this change will not propagate the labels/annotations to the descendants of the subnamespace, which we should definitely do.

So here's how I think we need to change this feature:
* In the [`SubnamespaceAnchor` CRD](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/api/v1alpha2/subnamespace_anchor.go#L67), add a new `.spec.labels` and `.spec.annotations` field, mirroring [the ones in `HierarchyConfiguration`](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/api/v1alpha2/hierarchy_types.go#L133).
* In the `HierarchyConfiguration` reconciler, modify the [`syncSubnamespaceParent` method](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/internal/hierarchyconfig/reconciler.go#L354) to copy the labels and annotations from the anchor to the `HierarchyConfiguration`. Replace anything that's there (e.g. don't attempt to add or merge). The rest of the reconciler will actually put the labels/annotations on the namespace and its descendants.
* Modify the validators to prevent adding illegal labels or annotations, at least as much as the current validators do (e.g. [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/internal/hierarchyconfig/validator.go#L527)). This may require some refactoring to avoid duplicating code. It's also ok to do this in a followup PR but please file an issue to do so.

Thanks!

/hold

mishamo:
Hi. Thanks for your feedback. It all sounds quite reasonable. Will look at updating this.

Regarding the use case for this feature, we would simply like to create child namespaces with labels and annotations. We can already do this by creating the namespace with a plain `Namespace` resource, then using `kubectl hns set` to set the parent. It feels like it would be much more elegant to be able to apply a single resource in a declarative manner, that could already contain the namespaces and annotations; in this case a `SubnamespaceAnchor`.

adrianludwin:
The other option is to create the NS resource, and then create the
HierarchyConfiguration resource inside it, setting both the parent and
labels/annotations simultaneously. It's still a two-step process, but at
least it doesn't have an imperative command as a part of it. But I'm happy
to make the changes to SubnamespaceAnchor as well if you feel it's
important.

mishamo:
Hi,

We've taken a look at the proposed implementation/UX and think there might be a slight mismatch in our usecase/goal vs the proposed approach.

We're looking to provide a mechanism for creating labels and annotations on namespaces that are created as a direct result of creating a `SubnamespaceAnchor`. This is a separate requirement to being able to propagate those annotations and labels. A possible approach could be to add `UnmanagedLabels` and `UnmanagedAnnotations` to `SubnamespaceAnchor.Spec`. These would then be copied across to the namespace that's created, but not propagated or managed by HNC. We could also add validation to make sure there are no clashes between managed and unmanaged labels and annotations.

Does that make sense? Or would there be a preferred approach to achieve this?

erikgb:
My main concern with allowing `SubnamespaceAnchor` to control labels/annotations on namespaces is the security aspect. AFAIK `SubnamespaceAnchor` exists [to allow unprivileged users to manage namespaces](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/how-to.md#administer-who-has-access-to-hnc-properties), which would normally be forbidden in a multi-tenant environment. With the new managed labels/annotations feature, the _cluster-admin_ can delegate permission to the _namespace-admin_ to set a **limited set of namespace labels/annotations** on the `HierarchyConfiguration` resource. IMO allowing unprivileged users to set namespace labels/annotations is not correct.

adrianludwin:
I think I'm ok to have anchors include _managed_ labels/annotations, since the admin of the parent namespace can always override  what's allowed (even if we don't have a great UX for that yet). I'm much less comfortable including _unmanaged_ labels/annotations on the subnamespaces for the reason that @erikgb says.

Can you give us a very concrete idea of why you need this feature? E.g.:
* Who is the cluster admin (or, at least, the admin of the namespace in which the anchors are being added)?
* Who is trying to create an anchor and why? What are they planning on doing with the resulting namespaces?
* What are the labels/annotations you're adding, and what do they mean? Why can't they be managed labels/annots?

This should help us either find another way for you to solve your problem or give us more clarity on how to implement this feature.

mishamo:
Hi. I've discussed this with my colleagues and we feel that you're right in terms of security.

We are the admins of the cluster and we have ~100 tenants. We would like generate parent namespaces for each tenant (with their individual roles and rolebindings to propagate to their children). Following that, tenants would be able to create their own children by applying `SubnamespaceAnchor` resources to their respective parent namespace.

We would like tenants (and ourselves, when generating the parents) to be able to add managed labels and annotations in one step, by creating the `SubnamespaceAnchor`. The proposal of adding a `Spec` to the `SubnamespaceAnchor` CRD and copying the labels and annotations across to the `HierarchyConfiguration` appears reasonable to us at face value. However, we appear to have stumbled upon some issues when implementing.

We've made some changes here: https://github.com/mishamo/hierarchical-namespaces/tree/changes

We have added a test to check for label propagation (which is very similar to the test in the hierarchyconfig/reconciler_test) here: https://github.com/mishamo/hierarchical-namespaces/blob/824376bcbdcd692af00e8a76134753c32bb2faeb/internal/anchor/reconciler_test.go#L136

The issue we're having is that when the labels in the anchor are updated for the parent namespace (foo), the reconcilers for the child namespace (bar) are not triggered, because nothing has changed in those namespaces. Thus the hierarchyconfig/reconciler only reconciles for foo, not bar. We could add logic to the anchor reconciler to update the hierachyconfiguration in the child, but that feels like crossing a responsibility boundary. We could also potentially delegate to reconcile children from parents (i.e. a direct call to `Reconcile` with the child namespace as an arg), but that feels quite intrusive and could have cascading side-effects.

We're scratching our heads a little as to the best approach. Any advice would be welcome.

adrianludwin:
If you change a label in the Hierarchy Config, it should trigger a reconciliation of all descendants (e.g. [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/internal/hierarchyconfig/reconciler.go#L525)). Is this not working for you?

adrianludwin:
Hmm, what exactly is the problem? E.g. imagine you have a namespace hierarchy like this:
```
foo
|-bar
```

Is the problem that you're changing the labels/annotations on the `bar` anchor in `foo`, and then the changes are not getting propagated to the `bar` namespace after the initial creation? If so, I think that would match [this logic](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/281b24ff4c6d9fbbed59f6ac473b9c3cad79450f/internal/hierarchyconfig/reconciler.go#L842), which says that when an anchor is changed, we re-reconcile the namespace the _anchor_ is in, not the namespace _created by_ the anchor. This is because of the finalization rules.

If that's the problem, the fix is pretty simple - just update the listening rules:

```
	anchorMapFn := func(obj client.Object) []reconcile.Request {
		return []reconcile.Request{
			{NamespacedName: types.NamespacedName{
				Name:      api.Singleton,
				Namespace: obj.GetNamespace(),
			}},
>>>			{NamespacedName: types.NamespacedName{
>>>				Name:      api.Singleton,
>>>				Namespace: obj.GetName(),
>>>			}},
		}
	}
```

This means that whenever an anchor is changed, we'll re-reconcile both the namespace that the anchor is in, as well as the subnamespace represented by that anchor.

mishamo:
Hi. Yes that sounds exactly like the issue. I'll take a look at your suggestion on Monday. Thanks.

adrianludwin:
Great, let me know!

mishamo:
Hi. The suggested fix appears to have resolved the issue we were having. We have pushed an update to the fork which contains changes that appear to work for us. There are some cases worth noting; once a SubnamespaceAnchor takes ownership of labels and/or annotations, then editing them in the HierarchyConfiguration has no effect (as reconciling simply overwrites them). We have added this behaviour to the test cases.

internal/objects/validator.go
```diff
@@ -356,3 +358,46 @@ func (v *Validator) InjectDecoder(d *admission.Decoder) error {
...
+		} else if !config.IsManagedAnnotation(a.Key) { // Only validate managed if key is valid
+			fldErr := field.Invalid(fldPath, a.Key, "not a managed annotation and cannot be configured")
+			allErrs = append(allErrs, fldErr)
+		}
+		// No validation of annotation values; only limitation seems to be the total length (256Kb) of all annotations
+	}
+	return allErrs
+}
+
+func ValidateMetaKey(k string, path *field.Path) *field.Error {
```
erikgb, line 391:
```suggestion
func validateMetaKey(k string, path *field.Path) *field.Error {
```

internal/objects/validator.go
```diff
@@ -356,3 +358,46 @@ func (v *Validator) InjectDecoder(d *admission.Decoder) error {
...
+	if errs := validation.IsQualifiedName(k); len(errs) != 0 {
+		return field.Invalid(path, k, strings.Join(errs, "; "))
+	}
+	return nil
+}
+
+func ValidateLabelValue(k, v string, path *field.Path) *field.Error {
```
erikgb, line 398:
```suggestion
func validateLabelValue(k, v string, path *field.Path) *field.Error {
```

config/webhook/manifests.yaml
```diff
@@ -38,60 +38,60 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /validate-hnc-x-k8s-io-v1alpha2-subnamespaceanchors
```
erikgb, line 41: It seems like the order is not consistent. What might be the reason? Another version of controller-gen? If so, I would prefer to re-generate this is in a separate PR.

erikgb (changes requested):
I did an initial scan. Looks promising.

/ok-to-test

config/webhook/manifests.yaml
mishamo: This reordering was caused by adding functionality to the `objects` validator. By moving the common validation logic elsewhere (e.g. `forest`), this will revert to the old order.

internal/objects/validator.go
```diff
@@ -356,3 +358,46 @@ func (v *Validator) InjectDecoder(d *admission.Decoder) error {
 	v.decoder = d
 	return nil
 }
+
+func ValidateManagedMeta(labels []api.MetaKVP, annotations []api.MetaKVP) field.ErrorList {
```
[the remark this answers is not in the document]
mishamo: We've moved this to `forest` as suggested

mishamo:
@adrianludwin How does this look to you? Is it likely that this would fit into the 1.0 release?

api/v1alpha2/subnamespace_anchor.go
```diff
@@ -68,9 +68,22 @@ type SubnamespaceAnchor struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	Spec   SubnamespaceAnchorSpec   `json:"spec,omitempty"`
 	Status SubnamespaceAnchorStatus `json:"status,omitempty"`
 }
 
+type SubnamespaceAnchorSpec struct {
+	// Lables is a list of labels and values to apply to the current namespace and all of its
```
adrianludwin, line 76: s/namespace/subnamespace/

api/v1alpha2/subnamespace_anchor.go
```diff
@@ -68,9 +68,22 @@ type SubnamespaceAnchor struct {
...
+	// descendants. All label keys must match a regex specified on the command line by
+	// --managed-namespace-label.
```
adrianludwin, line 78: "All label keys must be managed labels (see HNC docs) and must match a regex..."

Similar comments for annotations

internal/anchor/reconciler.go
```diff
@@ -380,6 +380,7 @@ func (r *Reconciler) getNamespace(ctx context.Context, nm string) (*corev1.Names
 func (r *Reconciler) writeNamespace(ctx context.Context, log logr.Logger, nm, pnm string) error {
 	inst := &corev1.Namespace{}
 	inst.ObjectMeta.Name = nm
+
```
adrianludwin, line 383: Remove please (since this is the only change in this file)?

internal/anchor/reconciler_test.go
```diff
@@ -2,15 +2,13 @@ package anchor_test
...
-	"testing"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/types"
-
 	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
 	"sigs.k8s.io/hierarchical-namespaces/internal/config"
 	. "sigs.k8s.io/hierarchical-namespaces/internal/integtest"
+	"testing"
```
adrianludwin, line 11: Can you please restore this part? All our Go files have three "paragraphs" of imports: the standard library, then external packages, then other HNC packages.

internal/forest/validator.go
```diff
@@ -0,0 +1,57 @@
+package forest
```
adrianludwin, line 1: Any reason not to put this in `config`? It doesn't seem to have anything to do with the rest of the `forest` package, but it does call into `config`.

internal/forest/validator.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
+	"sigs.k8s.io/hierarchical-namespaces/internal/config"
+	"strings"
```
adrianludwin, line 8: Split into three paragraphs (std, external, internal):

```
	"strings"

	"k8s.io/apimachinery/pkg/util/validation"
	"k8s.io/apimachinery/pkg/util/validation/field"

	api "sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
	"sigs.k8s.io/hierarchical-namespaces/internal/config"
```

internal/hierarchyconfig/reconciler.go
```diff
@@ -189,7 +189,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log logr.Logger, nm string)
 	r.updateFinalizers(log, inst, nsInst, anms)
 
 	// Sync the Hierarchy singleton with the in-memory forest.
-	needUpdateObjects := r.syncWithForest(log, nsInst, inst, deletingCRD, anms)
+	needUpdateObjects := r.syncWithForest(ctx, log, nsInst, inst, deletingCRD, anms)
```
adrianludwin, line 192: I've generally been trying to ensure that there are no calls to the client in `syncWithForest` because the mutex is being held and I want to be certain we don't make a call to the apiserver. I've recently learned that the caching is probably more guaranteed than I thought but I'm not certain, so I think I'd like to preserve this behaviour for now.

Instead, if you look up a few lines, you'll see the comment `Get a list of subnamespace anchors from apiserver`. We could move the code from`syncSubnamespaceParent` to here, and just pass the anchor (if it exists) down into `syncWithForest`. It's a bit messy but I'd prefer that for architectural reasons for now.

```
anchor, err := r.getSubnamespaceAnchorInstance(ctx, nsInst.Annotations[api.SubnamespaceOf], nm)
if err != nil && !errors.IsNotFound(err) {
  return err
}
```

This actually probably allows you to get rid of the loop in `syncSubnamespaceParent` as well, since we'll already have checked the apiserver for the anchor and will know if it's missing.

internal/hierarchyconfig/reconciler.go
```diff
@@ -846,6 +855,10 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, maxReconciles int) error
 				Name:      api.Singleton,
 				Namespace: obj.GetNamespace(),
 			}},
+			{NamespacedName: types.NamespacedName{
```
adrianludwin, line 858: Update the comment on line 851? And maybe add comments about why both are needed:
* The owner of the anchor needs to be reconciled to add or remove finalizers (IIRC)
* The subnamespace created by the anchor needs to be reconciled in case managed labels or annotations have changed

adrianludwin (commented):
This is looking great, thanks for making these changes. Just a few minor revisions.

/ok-to-test

adrianludwin:
E.g. see this commit message: https://github.com/kubernetes-sigs/hierarchical-namespaces/commit/48df98451954410686e7c6568116574ceee043db

internal/hierarchyconfig/reconciler.go
mishamo: Hi. I've pushed up changes based on this. All the tests still pass and the logic still makes sense to me. Would be good if you could confirm the changes are what you expected.

adrianludwin:
lgtm after one last cleanup of the commit message. It currently looks like you've merged a bunch of commit messages together? If you could modify it to look more like 48df98451954410686e7c6568116574ceee043db I'd appreciate that!

mishamo:
Thanks for that! I've updated the commit message with a brief description of the feature and removed the multiple interim commit messages.

mishamo:
I'm not quite sure why these tests are failing? The imports look OK to me...

erikgb:
> I'm not quite sure why these tests (go vet) are failing? The imports look OK to me and I can't reproduce this locally.

@mishamo I think the pipeline runs the result of adding your commit on top of the `master` branch. If you rebase on top on `master` you will be able to reproduce locally. I just did from a copy of your PR branch. 😄

mishamo:
@adrianludwin what are the chances of this making it into the 1.0 cut?

adrianludwin:
Let's get #160 in first, then we'll merge this. I think v1.0 will go out shortly after that.

adrianludwin (commented):
Thanks for this change!

/lgtm
/approve

rjbez17:
@adrianludwin I think we can remove this hold correct?

adrianludwin:
Whoops, thanks!

/hold cancel

PR #153, opened 2022-03-06
Introduce manager caching of Unstructured

erikgb:
Caching of Unstructured is disabled by default in controller-runtime, and that is suboptimal for HNC which is reconciling generic objects configured by users.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -224,6 +224,7 @@ func createManager() ctrl.Manager {
 	// it turns out to be harmful.
 	cfg.Burst = int(cfg.QPS * 1.5)
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		NewClient:              config.NewCachingClient,
 		Scheme:                 scheme,
 		MetricsBindAddress:     metricsAddr,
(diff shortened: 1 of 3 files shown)


adrianludwin:
Also what do you think the risk of this is? Should we delay it to v1.1?

erikgb:
> Also what do you think the risk of this is? Should we delay it to v1.1?

Hmmm, hard to tell. 🤔 We have been caching Unstructured for a long time in our controllers, and there seems to be quite performance gain. But on the other hand, the controller memory consumption might increase. How much depends on the cluster since and how many "object" resource types are configured in HNC. But resources tuning should any cluster-admin be able to do....

I can start by pulling out the "fixes" into a separate PR that can be merged. And leave the Unstructured caching (this PR) open for now?

adrianludwin:
Yeah let's punt this to v1.1.

rjbez17:
/hold 

for 1.1

erikgb:
I have now rebased this and fixed commit message and PR description. So should be ready for review - even if we decide to wait for 1.1 with this.

/cc @rjbez17

adrianludwin (commented):
/lgtm
/approve

PR #154, opened 2022-03-16
Replace unstructured query with structured query

erikgb:
Converted query for list of SubnamespaceAnchors to a structured query (no need to use UnstructuredList, and also had the list-kind bug described in the following).

Also fixing a bug in query for generic objects using UnstructuredList; when querying a List, one should use the object list kind (and not the object kind).

Tested: Ran unit-tests ('make test') successfully. After the controller probes PR got merged, I am sadly no longer able to run integration test ('make test-e2e'). The pod is just crash looping. :-(

Extracted from https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/153.


rjbez17:
Since we are delaying #153 to 1.1, should this wait as well?

erikgb:
> Since we are delaying #153 to 1.1, should this wait as well?

No, this is of an improvement/bugfix, and should go in. I have no idea why Unstructured was used for a CRD defined in HNC.

adrianludwin:
Agreed, I believe this is very safe.

adrianludwin:
Let's hold this until #160 goes in and we get a clean postsubmit run.

/hold

erikgb:
Rebased and tested on my workstation now. Should be GTM.

adrianludwin:
Postsubmits are clean again.
/hold cancel
/approve
/lgtm

PR #155, opened 2022-03-17
Remove last callers of webhooks.Deny function and delete it

erikgb:
Some error messages are slightly modified, in particular one error message that were formatted across multiple lines.

Tested: Ran both unit-tests ('make test') and integration test ('make test-e2e') successfully. ~~After the controller probes PR got merged, I am sadly no longer able to run integration test ('make test-e2e'). The pod is just crash looping. :-(~~

### DIFF
--- a/internal/objects/validator.go
+++ b/internal/objects/validator.go
@@ -54,6 +54,16 @@ type request struct {
 	obj    *unstructured.Unstructured
 	oldObj *unstructured.Unstructured
 	op     k8sadm.Operation
+	gvr    metav1.GroupVersionResource
+}
+
+func (req request) gr() schema.GroupResource {
+	return schema.GroupResource{Group: req.gvr.Group, Resource: req.gvr.Resource}
+}
+
+func (req request) name() string {
+	name := types.NamespacedName{Namespace: req.obj.GetNamespace(), Name: req.obj.GetName()}
+	return name.String()
 }
 
 func (v *Validator) Handle(ctx context.Context, req admission.Request) admission.Response {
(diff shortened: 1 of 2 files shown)


internal/objects/validator.go
```diff
@@ -54,6 +54,15 @@ type request struct {
...
 	op     k8sadm.Operation
+	gvr    metav1.GroupVersionResource
+}
+
+func (req request) GroupResource() schema.GroupResource {
+	return schema.GroupResource{Group: req.gvr.Group, Resource: req.gvr.Resource}
+}
+
+func (req request) NamespacedName() types.NamespacedName {
+	return types.NamespacedName{Namespace: req.obj.GetNamespace(), Name: req.obj.GetName()}
```
adrianludwin, line 65: I'd optimize these names for conciseness since they're internal to this file, e.g.

```
func (r request) gr() schema.GroupResource() {...}
func (r request) name() string {return ...}
```

This will make the boilerplate code much easier to read, e.g.

`return webhooks.DenyConflict(req.gr(), req.name(), err)`

instead of

`return webhooks.DenyConflict(req.GroupResource(), req.NamespacedName().String(), err)`

In addition, lowercase names on the functions will clearly indicate to the reader that these are local to this package. I often find that exported names on an unexported type like `req` are always a bit confusing to the reader.

adrianludwin (commented):
lgtm after one change. But let's hold until #160 is in and you can test this yourself.

adrianludwin:
/hold

until #160 is ready

internal/objects/validator.go
erikgb: Agreed and fixed! PTAL!

internal/objects/validator.go
```diff
@@ -54,6 +54,15 @@ type request struct {
...
 	oldObj *unstructured.Unstructured
 	op     k8sadm.Operation
+	gvr    metav1.GroupVersionResource
+}
+
+func (req request) gr() schema.GroupResource {
+	return schema.GroupResource{Group: req.gvr.Group, Resource: req.gvr.Resource}
+}
+
+func (req request) name() types.NamespacedName {
```
adrianludwin, line 64: Maybe just have this return a string rather than having to add `.String()` to every call?

erikgb: I prefer strongly typed functions, but I will fix. 😄

adrianludwin: I'd typically agree but for an unexported utility function, I prefer readability. If we were to export this, I'd also consider something more strongly typed.

adrianludwin (commented):
/lgtm
/approve

PR #156, opened 2022-03-19
change error message for check-fmt

### LINKED ISSUE #131: gofmt presubmit should be friendlier
Our presubmit should:
* Tell users to call `make fmt`, not `go fmt` (we don't want to reformat vendor files)
* Print out the diff so users can see what changes to make

rahulii:
Fixes #131 
Signed-off-by: rahulii <r.sawra@gmail.com>


rahulii:
cc: @adrianludwin

rjbez17:
/lgtm
/approve

Thanks!

Issue #159, opened 2022-03-24
Build version of hnc-manager.yaml manifest with cert-controller support


adrianludwin:
Cert-manager support is currently disabled [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/a34e5767c95fadfee51dd4e0aab23aa4c80d6677/config/default/kustomization.yaml#L20), since it's incompatible with the builtin cert-controller. Since lots of orgs like using cert-manager, we should probably make it easy to do so.

We should start building `hnc-manager-cc.yaml` and `hnc-manager-cm.yaml` to make it very easy to install HNC in either configuration.

/cc @erikgb 
/cc @rjbez17 
/good-first-issue

adrianludwin:
Fixed by #176, #177

/close

PR #160, opened 2022-03-24
Fix probes when installing on a fresh cluster

### LINKED ISSUE #158: Getting error "tls.crt: no such file or directory" when building HNC from main branch
Hi,

Environment details:
Rancher: 2.6.2
k8s cluster: EKS
k8s version: v1.21.5

We tried to generate manifests and image by building source from main branch. We were able to do that. However , when we deploy the manifests, we were getting error 
```
"msg":"problem running manager","error":"open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"runtime.main\n\t/usr/local/go/src/runtime/proc.go:255" 
```
The similar issue was mentioned https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/139#issuecomment-1036615144.

We would like to use HNC feature of label and annotation propagation and require to fix above error. It would be great if you could elaborate to set up the certificates using cert manager. We do have cert manager installed in our environment.

Thanks in advance.

adrianludwin:
Fixes #158

If the Secret with the certs doesn't exist, starting the
webhook server will fail and HNC will exit. Before we configured the
probes, we carefully didn't start the webhook server until after the
certs were ready, but as a side effect of creating the probe functions,
we accidentally started the webhook server before the certs were
generated. This worked fine on a cluster that already had the certs
(like the one I was testing on) but failed for everyone else (oops).

The fix is simply to use a non-default checker that knows whether the
certs have been generated, and avoids accidentally starting the webhook
server if the certs don't exist yet.

Tested: manually on a cluster without the Secret. Without this change, I
can see the error from the webhook server complaining that the certs
don't exist; with this change, I can see that that in HNC's first
invocation, we get some "healthz check failed" messages, but the cert
controller runs, generates the certs, and restarts HNC. The second
invocation works just fine.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -16,7 +16,9 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"flag"
+	"net/http"
 	"os"
 	"strings"
 
@@ -90,7 +92,6 @@ func main() {
 	metricsCleanupFn := enableMetrics()
 	defer metricsCleanupFn()
 	mgr := createManager()
-	setupChecks(mgr)
 
 	// Make sure certs are generated and valid if webhooks are enabled and internal certs are used.
 	setupLog.Info("Starting certificate generation")
@@ -100,9 +101,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupProbeEndpoints(mgr, certsReady)
+
 	// The call to mgr.Start will never return, but the certs won't be ready until the manager starts
-	// and we can't set up the webhooks without them. So start a goroutine which will wait until the
-	// certs are ready, and then create the rest of the HNC controllers.
+	// and we can't set up the webhooks without them (the webhook server runnable will try to read the
+	// certs, and if those certs don't exist, the entire process will exit). So start a goroutine
+	// which will wait until the certs are ready, and then create the rest of the HNC controllers.
 	go startControllers(mgr, certsReady)
 
 	setupLog.Info("Starting manager")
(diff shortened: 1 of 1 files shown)


erikgb:
While this solution seems ok, I think the canonical solution would be to start with a dummy certificate that later will be replaced by the certificate-provider. Just to solve the bootstrapping problem. That is how I would solve this using cert-manager. But I also think certificates should be provided **to** the application, and _not provided by_ the application.

/lgtm

erikgb:
Oooops, I didn't think my lgtm would merge this. Sorry! I should have added a hold to get some discussions around my comment.

rjbez17:
I wonder if it may be worthwhile to reconsider removing the cert creation logic altogether from startup? Just have a dependency that certs are provided. We take in a secret and k8s will prevent us from starting until the secret exists. Then we can have a separate (optional) startup job that does any and all cert creation logic or leave it to users to not run the job if they have their own solution (cert manager). I know we discussed this in the past but I don't remember why we didn't like that approach? To handle cert rotations the job could be a cron job for self signed and cert manager handles for others. We just need a watcher on the mounted secret file to reload HNC on change.

This would guarantee we have a cert at startup.

adrianludwin:
I really like having a one-step install to make it easy to try out, not to
mention write tests for or even just test manually on a new cluster. Plus
cert-controller already existed so we didn't need to write anything
ourselves.

I'd rather go with the approach in #159 where we simply build multiple
manifests to make it easy for cert-manager users as well. wdyt?

erikgb:
I agree with @rjbez17. The current approach will complicate the code more over time, and put HNC further away from controller-runtime and kubebuilder. I'll vote for cert-manager as the default - as suggested by kubebuilder in the skaffolding process. And eventually an opt-in for cert-rotator, that could be used in tests - with some test adjustments. Or just use a self-signed cert in the tests and drop the cert-rotator.

adrianludwin:
We did start with cert-manager as the default (and only) method but it made
development and experimentation difficult.

Can you or Ryan maybe create a setup script to show what a standalone
installer would look like, without either cert-controller or cert-manager?

PR #161, opened 2022-03-24
Install Kind via 'go install'

adrianludwin:
Postsubmits and periodics are failing with a message to use 'go
install', so trying that. Also upgrading to lastest Kind version at the
same time.

Tested: none, no real way to test without submitting, plus all the tests
are currently broken anyway so can't make things worse.


adrianludwin:
Hi, may I get a quick lgtm on this please?

/assign @erikgb 
/assign @rjbez17

adrianludwin:
Postscript: postsubmits are running correctly after this change.

PR #167, opened 2022-03-28
Introduce manifest variants

### LINKED ISSUE #151: Document how to deploy HNC without its webhooks
Webhooks are dangerous and may not be necessary for all users.

/assign me

adrianludwin:
This change updates the config and Makefile to produce multiple variants
of the HNC manifests. The initial list is "default" (same as today),
"default-cm" (same, but configured for cert-manager instead of the
internal cert-controller), and ha (replicated webhooks, using the
internal cert controller).

Also made some updates to the Prow tests to allow for variant testing,
as well as upgrading the ability to test the Prow tests locally.

Tested: verified that default.yaml is the same as the prior
hnc-manager.yaml, apart from some reordered command line args. Verified
that smoke tests work on ha.yaml. Did not do _any_ testing on the
cert-manager variants. Locally ran the prow e2e tests on both the HA and
default configs. Also did an experimental release from my personal repo
and bumped up the timeout since we've added more build targets.

Fixes #151


adrianludwin:
To see the output of this process, see `default.yaml` and `ha.yaml` at https://github.com/adrianludwin/hierarchical-namespaces/releases/tag/v1.0.0-test3.

config/variants/default-cm/kustomization.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+patchesStrategicMerge:
+- webhookcainjection_patch.yaml
+
+vars:
+- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
+  objref:
+    kind: Certificate
+    group: cert-manager.io
+    version: v1alpha2
```
erikgb, line 26:
```suggestion
    version: v1
```

config/variants/default-cm/kustomization.yaml
```diff
@@ -0,0 +1,47 @@
...
+    name: serving-cert # this name should match the one in certificate.yaml
+  fieldref:
+    fieldpath: metadata.namespace
+- name: CERTIFICATE_NAME
+  objref:
+    kind: Certificate
+    group: cert-manager.io
+    version: v1alpha2
```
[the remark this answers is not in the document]
adrianludwin: Deon

config/variants/default-cm/kustomization.yaml
adrianludwin: Done

PR #168, opened 2022-03-29
Validate subnamespace anchors on update

adamdougal:
Now that labels and annotations can be added to subnamespace anchors, these should now also be validated when they are updated to avoid invalid labels and annotations being added after creation.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -48,6 +48,7 @@ webhooks:
     - v1alpha2
     operations:
     - CREATE
+    - UPDATE
     - DELETE
     resources:
     - subnamespaceanchors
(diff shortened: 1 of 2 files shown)


adrianludwin:
Hey @adamdougal , thanks for this PR. It lgtm, can you please just make a few last changes?

1. Squash all commits into a single commit
2. Add a "Tested:" comment to that commit, similar to this one: https://github.com/kubernetes-sigs/hierarchical-namespaces/commit/0f29b00b9f5427f8e6a5d11e401f5ec0168df2b3

Then we'll get this approved and checked in. Thanks!

adrianludwin:
Also, do you think this should be backported to v1.0.1? It seems reasonable to me.

erikgb:
> Also, do you think this should be backported to v1.0.1? It seems reasonable to me.

I think so, it might even fix https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/175? At least improve the UX considerably.

/ok-to-test

erikgb:
/lgtm

I also suggest cherry-picking this to the v1.0 branch.

adrianludwin:
/lgtm
/approve

Thanks for this! Can you please open a second PR against the v1.0 branch?

PR #173, opened 2022-03-29
Add readiness probe

### LINKED ISSUE #170: HNC slow to become Ready on 1.21 and earlier
I've found that HNC becomes Ready on 1.22 after about 10s, while on 1.20 and 1.21 it takes about 80+ seconds. Adding a Readiness probe seems to solve the problem - I'm not sure why, since "no readiness probe" is supposed to be equivalent to "always ready" and this does seem to be the case in 1.22. But I can't find any evidence of a bug that got fixed in 1.22 and I'm surprised that something this obviously wrong would have survived so long.

Either way, the solution seems pretty straightforward and I'll make the fix today.

/cc @erikgb

adrianludwin:
On GKE 1.21 and earlier, I noticed HNC taking a long time (~80s) to
become Ready (for more details, see #170). Adding a readiness probe
fixes the problem.

Tested: before this change, on GKE 1.20 and 1.21, I manually see HNC
taking a long time to start, and the e2e tests that require reinstalling
HNC fail because it the deadlines are exceeded. With this change, I can
see HNC becoming ready in ~10s on GKE 1.20 and all the e2e tests pass.

Fixes #170

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -67,6 +67,11 @@ spec:
             port: 8081
           failureThreshold: 1
           periodSeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8081
+          periodSeconds: 5
         startupProbe:
           httpGet:
             path: /readyz


adrianludwin:
Whoops, I accidentally merged the v1.0 PR to master instead of `v1.0`. Can you please re-approve? Sorry about that!

/assign @erikgb 
/assign @rjbez17

PR #176, opened 2022-03-30
Fix cert-manager build error

adrianludwin:
Tested: none, I have no idea if this works or not. For experimentation
purposes only.


adrianludwin:
/assign @erikgb 

Maybe try this fix out before approving? If it works I'll put this in v1.0.x.

erikgb:
Verified that this fixes the build error, and `manifests/default-cm.yaml` is generated. I will also check if the installation with cert-manager works, but it is probably better if I submit a dedicated PR to fix any issues.

/lgtm

PR #180, opened 2022-04-04
Decouple reconcilers from each other

adrianludwin:
Instead of having the HC reconciler directly enqueue other reconcilers,
it now has a level of indirection through the forest. This allows the
anchor and object reconcilers - and, soon, the Hierarchical Resource
Quota reconcilers - to be notified of changes to the hierarchy, without
the HC reconciler having to know about them. (This change was prompted
by my work on HRQs, which I felt would overly complicate the HC
reconciler).

At the same time, I've removed the enqueues of the HNC Config
reconciler. Instead of reconciling whenever there was a change, with a
rate-limited maximum, it's now just on a periodic ticker.

Finally, I used the SetupWithManager functions as constructors to take
the initialization of certain fields out of internal/setup and put them
into the per-feature packages.

Tested: all e2e tests still work (I didn't try the ones that require
repair but those should be unaffected). Also tested this with my HRQ
changes and HRQ works correctly and fits in with minimal changes to the
rest of HNC.


adrianludwin:
/assign @rjbez17 

Also CC @erikgb , please feel free to leave comments but I'd like Ryan to approve since he has a bit more context here.

The HRQ implementation I allude to in the commit description is at https://github.com/adrianludwin/hierarchical-namespaces/tree/hrq.

internal/objects/reconciler.go
```diff
@@ -108,21 +104,50 @@ type HNCConfigReconcilerType interface {
 //
 // +kubebuilder:rbac:groups=*,resources=*,verbs=*
 
-// SyncNamespace can be called manually by the HierarchyConfigReconciler when the hierarchy changes.
-// It enqueues all the current objects in the namespace and local copies of the original objects
-// in the ancestors.
-func (r *Reconciler) SyncNamespace(ctx context.Context, log logr.Logger, ns string) error {
+// OnChangeNamespace is called whenever the a namespace is changed (according to
```
rjbez17, line 107: nit: the a

internal/objects/reconciler.go
```diff
@@ -108,21 +104,50 @@ type HNCConfigReconcilerType interface {
...
+			if err == nil {
+				break
+			}
+			log.Error(err, "Couldn't list objects")
+			time.Sleep(10 * time.Second)
+			if attempts == 3 {
+				// We can get into a permanent failure loop if, for example, the GVK has been deleted from
+				// the server (i.e. it was a CRD). Rather than looping forever, let's restart and hope that
+				// fixes it.
+				log.Info("Couldn't list objects for 30s. Something's gone very wrong; hopefully a restart will fix it.")
```
rjbez17, line 133: My concern here is that when we hit this, HNC will restart, which means the webhooks wont respond. If the api server is under heavy load (which I've seen a lot in the past on new cluster bring ups), it could lead us into weird loop of HNC being down, so webhooks fail, so things don't clear from api server queues, so HNC has issues listing objects. 

I don't think it is a huge concern, but I have had to temporarily delete webhooks for HNC, Kyverno, OPA, etc in the past due to some of these issues to get things to recover, which is never ideal. Do you think it's worthwhile to expose MaxRetryTimeout as a cli flag so it can be adjusted?

rjbez17 (commented):
A nit and a question, otherwise lgtm

internal/objects/reconciler.go
adrianludwin: Yeah maybe I should just take this out and see if it ever becomes a problem? E.g. the goroutine should still terminate but maybe give any _one_ goroutine a 60s timeout and then have it just exit.

We should probably look more holistically at things that would cause permanent failures in the object reconcilers, e.g. having a CRD deleted, and see how best to address that.

internal/objects/reconciler.go
adrianludwin: Fixed, thanks.

adrianludwin:
Oooh, that's a good idea. I'll add that to #182.

PR #196, opened 2022-04-12
Disable HRQ e2e tests

adrianludwin:
HRQ isn't on in the default manifests so we shouldn't try to test it.

Tested: verified that the right number tests are "pending" when I run
`make test-e2e`. Didn't wait to verify exactly which ones they were.


adrianludwin:
/assign @erikgb 
/assign @rjbez17 

May I get a quick lgtm for this? The nightly e2e tests are broken because of it. Thanks!

erikgb:
/lgtm

Thanks!

PR #197, opened 2022-04-12
Use in-tree metav1.Condition instead of custom copied type

erikgb:
This is a refactoring. K8s 1.20 is EOL, so it should be safe to use the master type from api-machinery.

/cc @adrianludwin 
/cc @rjbez17


internal/forest/namespaceconditions.go
```diff
@@ -50,12 +59,19 @@ func (ns *Namespace) ClearConditions() {
...
 	if root := ns.GetHaltedRoot(); root != "" && root != ns.name {
-		msg := fmt.Sprintf("Propagation paused in %q and its descendants due to ActivitiesHalted condition on ancestor %q", ns.name, root)
-		ret := []api.Condition{api.NewCondition(api.ConditionActivitiesHalted, api.ReasonAncestor, msg)}
+		ret := []metav1.Condition{{
+			Type:   api.ConditionActivitiesHalted,
+			Status: metav1.ConditionTrue,
+			// TODO(adrianludwin) Is the following statement really correct?
+			// Set time as an obviously wrong value 1970-01-01T00:00:00Z since we
+			// overwrite conditions every time.
+			LastTransitionTime: metav1.Unix(0, 0),
```
adrianludwin, line 70: I'd be happy to just leave that out since I assume it's the zero value anyway. Also SetStatusCondition seems to set it automatically if it's zero-valued anyway.

adrianludwin (commented):
One nit, but otherwise lgtm

/ok-to-test

internal/forest/namespaceconditions.go
erikgb: Hmmm, removing it seems to introduce failing tests. I will have to investigate a bit more or leave it as it is now. WDYT? There are two quite different use-cases calling this function: `loadNamespaceConditions` and `syncConditions`. I suspect we need two functions with a bit different semantics to fix the failing tests.

adrianludwin: ok maybe just update the comment (that this code doesn't work without this statement) and I'll approve?

adrianludwin (commented):
/lgtm
/approve

internal/forest/namespaceconditions.go
erikgb: Done as suggested. I think you could take a look. It seems to touch some of the core functions in HNC that I do not know very well, at least not yet.

PR #202, opened 2022-04-19
Fix manifest kustomization for metrics service

### LINKED ISSUE #115: Metrics svc port mismatch on the release/installation file
The target port on the metrics service should be `8080` or `metrics` instead of `https` as shown below. 

```yaml
  ports:
  - name: https
    port: 8443
    protocol: TCP
    targetPort: metrics
```

Since the k8s manifest has the `targetPort: https` the metrics are not scraped by the Prometheus as there isn't any port on that name specified in the service.

erikgb:
Also removed unused ClusterRole and ClusterRoleBinding that is supposed to grant access through kube-rbac-proxy, which is currently not in use in HNC.

Tested: Verified that metrics service works as annotated by prometheus service annotations

Fixes #115 

With this change, I can scrape metrics as described by metrics-service prometheus annotations:
````
$ kubectl -n hnc-system run -it --rm --restart=Never curl --image=curlimages/curl sh
If you don't see a command prompt, try pressing enter.
/ $ curl http://hnc-controller-manager-metrics-service:8080/metrics
````

Before this change, this fails:
````
$ kubectl -n hnc-system run -it --rm --restart=Never curl --image=curlimages/curl sh
If you don't see a command prompt, try pressing enter.
/ $ curl https://hnc-controller-manager-metrics-service:8443/metrics
curl: (7) Failed to connect to hnc-controller-manager-metrics-service port 8443 after 0 ms: Connection refused
/ $ 
````

Another option is to configure the kube-rbac-proxy. Please let me know if that solution is preferred.

/cc @adrianludwin

### CODE REVIEW

File: config/manager/manager.yaml [hunk 1]
```diff
@@ -40,9 +40,6 @@ spec:
         - /manager
         args:
         - "--webhook-server-port=9443"
-        # If /config/default/manager_auth_proxy_patch.yaml is used, the port in
-        # --metrics-addr must match the value of --upstream passed to
-        # kube-rbac-proxy.
         - "--metrics-addr=:8080"
```
rjbez17, line 43: Wouldn't just changing this to 8443 fix our issue and still allow us to use HTTPS? I'm not sure we want to take a step back in security and not use HTTPS for metrics.

erikgb: > Wouldn't just changing this to 8443 fix our issue and still allow us to use HTTPS? I'm not sure we want to take a step back in security and not use HTTPS for metrics.

No, there is nothing providing a HTTPS listener (on port 8443). Then we need to include the kube-rbac-proxy sidecar and route all metrics scrape request through its HTTPS listener - forwarding the request to the controller-manager metrics HTTP endpoint. That is an alternative mentioned in the PR description, though more complex, and I don't see the true benefit of this. I expect many K8s distros to block namespace ingress traffic by default.

adrianludwin: Yeah we took out kube-rbac-proxy a while ago (it has a bunch of problems from a supply chain perspective) and I'd rather not put it back in without good justification.

adrianludwin (commented):
/approve
/lgtm
/hold
/assign @rjbez17 

I think this is good unless Ryan has any remaining concerns.

config/manager/manager.yaml
rjbez17: We don't need rbac proxy for https, that is only needed for authz. We already serve https traffic for the webhook on 9443. So I guess I don't see why we couldn't also use it (or have the option) for 8443 for metrics. However, that is probably out of scope of this PR if metrics are currently broken. 

I created [this](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/204) to revisit this.

rjbez17:
I opened https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/204 to follow up on my concerns.

/hold cancel 
/lgtm

erikgb:
@rjbez17 @adrianludwin Thanks for accepting this PR. I think this is a bugfix, and should be cherry-picked to the `v1.0` branch. The metrics setup is borked in the 1.0 release. WDYT?

adrianludwin:
sgtm, I'd happily approve that and then release v1.0.1.

erikgb:
> sgtm, I'd happily approve that and then release v1.0.1.

Nice, here is the cherry-pick PR: https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/207

Issue #200, opened 2022-04-19
Proposal: remove anchor finalizers as soon as subns starts getting deleted


adrianludwin:
Today, if you delete a subnamespace anchor, the anchor is not fully deleted until the underlying subnamespace is deleted. Namespaces are usually deleted within several seconds, but can sometimes get "stuck," in which case the anchor isn't deleted either.

This behaviour has two problems:

- If the anchor isn't deleted, it can prevent the parent namespace from being deleted, since namespaces cannot be deleted until all the objects in them have also been deleted.
- The code used to implement this behaviour in HNC is quite complicated, and changing it may introduce subtle bugs.

As a result, we're proposing a simplification: HNC will let the anchor be deleted as soon as it starts deleting the subnamespace. This means that the anchor will typically vanish virtually "instantly" - that is, typically less than a second after you say `kubectl delete subns <foo>`. The subnamespace itself will be put into the "deleting" state and will be removed whenever Kubernetes is able to do so.

If you've never checked to see if an anchor has actually been deleted before, this change won't affect you. Otherwise, you need to be aware that simply because the anchor has disappeared, there's a chance that the underlying namespace could still be there, either because it's still being deleted, or because it's gotten stuck somehow.

Issue #142, opened 2022-02-02
Resources status should be compatible with kstatus


erikgb:
As a user of tools like [FluxCD](https://github.com/fluxcd/flux2), I would like Flux to know if/when my HNC resources are fully reconciled and operational (ready/healthy). Flux supports [health assessment](https://fluxcd.io/docs/components/kustomize/kustomization/#health-assessment) for all/most in-tree resource types, but it turns out that Flux also can support custom resources - if the resource expose a status compatible with [kstatus](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#readme).

I think this should be fairly easy to implement, and I would be happy to submit a PR to fix it. But we need to decide what to do with the current content of resource statuses? Can we delete something (breaking), or should we avoid breaking changes and keep duplicated status content with the same meaning?

adrianludwin:
I _think_ our `Condition` struct is compatible with the standard (see [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/22c4260b6aa751bc7a65f74d8d48cd9bd9a3be83/api/v1alpha2/hierarchy_types.go#L176) so that's a good start.

As for _replacing_ the existing conditions, the issue I have is that `HierarchyConfiguration` currently has one Condition that means "this entire namespace is in a bad state" (`ActivitiesHalted`) and another that means "something's gone wrong but it's not a disaster" (`BadConfiguration`). I think kstatus would lump these both into `Failed`, and I don't particularly want to lose the existing distinction. But I'm totally on board with _adding_ a kstatus `Stalled` condition if either of the above are true. (I'd rather not use the `Ready=true/false` condition since namespaces aren't exactly ever "ready").

Similarly, `HNCConfiguration` has three conditions: `BadConfiguration`, `OutOfSync` and the poorly named `NamespaceCondition` (there's a problem in a namespace in HNC). Similarly, I think kstatus would map the first two of these to `Stalled`. The third one is more informational and doesn't have anything to do with the `HNCConfiguration` objects itself so probably doesn't need a kstatus-compliant condition.

How does that sound?

erikgb:
I think it could make sense to keep the existing conditions when the HNC resources are not fully reconciled. AFAIK the most important requirements in kstatus seem to be:
- Add a new field: `status.observedGeneration` to be able to detect when the controller is aware of the resource state, and has started working with it.
[...]

PR #203, opened 2022-04-19
Refactor adding/removing/contains finalizers

erikgb:
To be able to distinguish between updating just the status and updating the full resource (in a follow-up PR). A controller should just update the status of the controlling resource, with some exceptions - i.e. finalizers. Currently HNC always updates the complete resource, and not just the status, and this has to be fixed to open up for #142.

Relates to #200.

### DIFF
--- a/internal/anchor/reconciler.go
+++ b/internal/anchor/reconciler.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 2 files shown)


adrianludwin:
I'm not really seeing how this change improves the _organization_ of the finalizers? It replaces some custom code with a slightly safe utility function but that's it, AIUI. I'm ok with the change (other than one caveat) but I'm wondering if I missed something in the review, or if this PR was supposed to include something that got lost somehow?

internal/anchor/reconciler.go
```diff
@@ -168,13 +171,16 @@ func (r *Reconciler) onDeleting(ctx context.Context, log logr.Logger, inst *api.
 		log.Info("Deleting subnamespace due to anchor being deleted")
 		return true, r.deleteNamespace(ctx, log, snsInst)
 	case r.shouldFinalizeAnchor(log, inst, snsInst):
-		log.V(1).Info("Unblocking deletion") // V(1) since we'll very shortly show an "anchor deleted" message
-		inst.ObjectMeta.Finalizers = nil
-		return true, r.writeInstance(ctx, log, inst)
+		if controllerutil.ContainsFinalizer(inst, api.MetaGroup) {
```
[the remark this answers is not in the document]
erikgb: Thanks! Fixed

erikgb:
Ok, take 2 ready. PTAL!

internal/anchor/reconciler.go
```diff
@@ -113,8 +114,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if inst.DeletionTimestamp.IsZero() {
+		// Add finalizers on all non-forbidden anchors to ensure it's not deleted until
+		// after the subnamespace is deleted.
+		if !controllerutil.ContainsFinalizer(inst, api.MetaGroup) {
+			controllerutil.AddFinalizer(inst, api.MetaGroup)
+		}
+	} else {
+		// Stop reconciliation as anchor is being deleted
+		return ctrl.Result{}, r.onDeleting(ctx, log, inst, snsInst)
 	}
```
adrianludwin, line 126: Can you please re-order the statements to have the early-exit come first? Also, `AddFinalizer` is idempotent so you don't need to do the check first.

```
	// Handle the case where the anchor is being deleted.
	if !inst.DeletionTimestamp.IsZero() {
		return ctrl.Result{}, r.onDeleting(ctx, log, inst, snsInst)
	}

	// Add finalizers on all non-forbidden anchors to ensure it's not deleted until
	// after the subnamespace is deleted.
	controllerutil.AddFinalizer(inst, api.MetaGroup)
```

adrianludwin (commented):
One last change and then lgtm.

internal/anchor/reconciler.go
erikgb: I will invert the if. Going to add a conditional update in my next PR, but I'll remove the conditional around `AddFinalizer` for now. :-)

erikgb: Fixed now, PTAL!

adrianludwin (commented):
/lgtm
/approve

PR #205, opened 2022-04-23
Remove obsolete validation TODOs

erikgb:
I added some TODOs when improving the error messages from webhook validation. Most of these were suggesting to review, and eventually change, the status code returned. I realize now that this is probably not feasible, and apologize for adding the TODOs in the first place.


adrianludwin (commented):
/lgtm
/approve

FYI I wouldn't bother to run integ tests for a change list this, the risk is minimal plus we have the postsubmits in case something crazy goes wrong.

PR #206, opened 2022-04-23
Read-only client in webhooks only mode

### LINKED ISSUE #195: Make HRQ HA-compliant
#183 doesn't support the ReadOnly fields needed to support HA deployments of HNC. We should either update the HRQ reconcilers to support ReadOnly, or else come up with a different overall solution (e.g. modify the K8s client so that all write operations are no-ops in read-only mode).

/assign

### LINKED ISSUE #198: Read-only mode implemented with read-only client
The newly introduced read-only mode to make HNC more highly available is currently implemented with a `ReadOnly` field on all reconcilers, and every write operation must guarded with a check on the value of this field. This is not robust enough IMO, and should be improved. Contributors not aware of this feature could easily forget adding this guard, leaving it up to reviewers to catch it - creating a troublesome contributor process with a high risk of errors being introduced in HNC. As an example, I dumped into the following code today, which misses this guard:

https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/c5662040fb67c72a22085f50653cf74a1317d5f2/internal/hierarchyconfig/reconciler.go#L141

I suggest to instead implement the read-only mode by supplying the reconcilers with a delegating client, where all writes are implemented as a no-op. A bit like controller-runtime does for their caching client.

erikgb:
This appears like a more robust implementation of the webhooks only mode than the current one. It ensures that the client provided to all components are not able to perform any write operations towards the API.

Tested: Ran both unit-tests ('make test') and integration test ('make test-e2e') successfully. The integration tests were verified both in default and HA mode.

Fixes #198, fixes #195


internal/config/client.go
```diff
@@ -0,0 +1,99 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
+)
+
+// NewCachingClient is an alternative implementation of controller-runtime's
+// default client for manager.Manager.
+// The only difference is that this implementation sets `CacheUnstructured` to `true` to
+// cache unstructured objects.
+func NewCachingClient(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
```
adrianludwin, line 17: Should this be an unexported function now?

internal/config/client.go
```diff
@@ -0,0 +1,99 @@
...
+func (c ReadOnlyClient) Scheme() *runtime.Scheme {
+	return c.client.Scheme()
+}
+
+func (c ReadOnlyClient) RESTMapper() meta.RESTMapper {
+	return c.client.RESTMapper()
+}
+
+type nopStatusWriter struct {
+}
```
adrianludwin, line 91: Nit: I usually prefer the one-line `type nopStatusWriter struct{}`

internal/config/client.go
```diff
@@ -0,0 +1,99 @@
...
+		return c, err
+	}
+}
+
+// ReadOnlyClient is a client decorator that ensures no write operations are possible.
+type ReadOnlyClient struct {
+	client client.Client
+}
+
+func (c ReadOnlyClient) GetReadWriteClient() client.Client {
```
adrianludwin, line 46: Maybe add a comment explaining why we have this function. Or can we factor it out instead? Pass a normal client to the validator that needs it, so we don't need this anymore?

Alternatively, `Create()` (line 58) could have special-case code for SubjectAccessReview objects?

adrianludwin (commented):
Awesome change :) Just a few suggestions

internal/config/client.go
erikgb: Yes, it can be unexported now. Just wonder why you want to export/unexport based on the currentusage? To me, it is more natural to export/unexport based on weather the symbol could/should be used outside the package.

internal/config/client.go
erikgb: I prefer to allow all components to use the same client. Was thinking about special handling for SAR in the `Create` func. I definitely prefer that over passing selective client around, and will give that a shot....

internal/config/client.go
adrianludwin: I generally default to "nothing exported unless needed." It makes it that much easier to understand what the entrypoints to a package are when I'm reading the code.

internal/config/client.go
adrianludwin: 👍

internal/config/client.go
erikgb: Okey!

erikgb:
@adrianludwin I have addressed all your comments. PTAL!

adrianludwin (commented):
I like it! Thanks!

/lgtm
/approve

Issue #16, opened 2021-05-05
Allow inclusive configuration of resources to propagate


yiqigao217:
<a href="https://github.com/hferentschik"><img src="https://avatars.githubusercontent.com/u/296354?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [hferentschik](https://github.com/hferentschik)**
_Tuesday Mar 16, 2021 at 09:34 GMT_
_Originally opened as https://github.com/kubernetes-sigs/multi-tenancy/issues/1436_

----

When synchronizing additional resource types, eg Secrets, it is currently only possible to configure this globally (see #1435) and the propagation configuration only allows to exclude certain resources from propagation (via annotation).

It would be nice to have an inclusive way of configuring propagation as well. For example, assume the root namespace contains multiple secrets and I would like to only propagate specific secrets (something like what [reflector](https://github.com/EmberStack/kubernetes-reflector) does). In this case I would like to annotate the Secret to propagate and maybe somehow configure to which child namespaces it should be propagated. 

This might require changing the propagation modes and adding a new one.

yiqigao217:
<a href="https://github.com/adrianludwin"><img src="https://avatars.githubusercontent.com/u/20559844?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [adrianludwin](https://github.com/adrianludwin)**
_Tuesday Mar 16, 2021 at 14:40 GMT_

----

Yes, this is a good idea and we considered it when we were first [designing exceptions](https://docs.google.com/document/d/17J8icBEDvLLoPT4kQ4ArZcCerRweDY-TpJ48DJKpHJ0):

> In addition, we will consider adding a global `AllowPropagate` mode to `HNCConfig`. This could be useful for types such as Secrets that we don’t want to always propagate. When the propagation mode is `Propagate`, objects will always be propagated unless an annotation is set; in `AllowPropagate`, the reverse is true and the objects will only be propagated if an annotation is set. If we do this, we’ll also a `propagate.hnc.x-k8s.io/all=true` annotation to allow an object to opt into full propagation.

We haven't seen a strong demand from this from our own customers, but I think the interface and semantics are fairly clear so I'd welcome PRs for this feature. Note that the majority of the work here would likely be in coming up with suitable integration tests; the code itself would likely be relatively easy.

/good-first-issue

yiqigao217:
[...]

PR #217, opened 2022-07-24
Inclusive configuration of resources to propagate

mzeevi:
Addresses #16. As suggested in the issue, an additional SyncornizationMode called `Allow` is added (and not `AllowPropagate` to allow the kubetcl plugin keep the same logic of converting the strings to Title).

Allow only allows propagation of objects when a selector is set. When a selector is set, `Allow` follows the same logic as `Propagate`. When no selectors are set, the object is not propagated.

In addition, an `all` selector is added.

New unit testing and e2e testing were made for this mode and for the new selector.

The documentation and quickstart testing are left untouched.

### DIFF
--- a/internal/hierarchyconfig/validator.go
+++ b/internal/hierarchyconfig/validator.go
@@ -246,11 +246,11 @@ func (v *Validator) getConflictingObjects(newParent, ns *forest.Namespace) []str
 	if newParent == nil {
 		return nil
 	}
-	// Traverse all the types with 'Propagate' mode to find any conflicts.
+	// Traverse all the types with 'Propagate' mode or 'AllowPropogate' mode to find any conflicts.
 	conflicts := []string{}
 	for _, t := range v.Forest.GetTypeSyncers() {
-		if t.GetMode() == api.Propagate {
-			conflicts = append(conflicts, v.getConflictingObjectsOfType(t.GetGVK(), newParent, ns)...)
+		if t.CanPropagate() {
+			conflicts = append(conflicts, v.getConflictingObjectsOfType(t.GetGVK(), t.GetMode(), newParent, ns)...)
 		}
 	}
 	return conflicts
@@ -258,7 +258,7 @@ func (v *Validator) getConflictingObjects(newParent, ns *forest.Namespace) []str
 
 // getConflictingObjectsOfType returns a list of namespaced objects if there's
 // any conflict between the new ancestors and the descendants.
-func (v *Validator) getConflictingObjectsOfType(gvk schema.GroupVersionKind, newParent, ns *forest.Namespace) []string {
+func (v *Validator) getConflictingObjectsOfType(gvk schema.GroupVersionKind, mode api.SynchronizationMode, newParent, ns *forest.Namespace) []string {
 	// Get all the source objects in the new ancestors that would be propagated
 	// into the descendants.
 	newAnsSrcObjs := make(map[string]bool)
@@ -266,7 +266,7 @@ func (v *Validator) getConflictingObjectsOfType(gvk schema.GroupVersionKind, new
 		// If the user has chosen not to propagate the object to this descendant,
 		// then it should not be included in conflict checks
 		o := v.Forest.Get(nnm.Namespace).GetSourceObject(gvk, nnm.Name)
-		if ok, _ := selectors.ShouldPropagate(o, o.GetLabels()); ok {
+		if ok, _ := selectors.ShouldPropagate(o, o.GetLabels(), mode); ok {
 			newAnsSrcObjs[nnm.Name] = true
 		}
 	}
--- a/internal/kubectl/configset.go
+++ b/internal/kubectl/configset.go
@@ -27,8 +27,8 @@ import (
 )
 
 var setResourceCmd = &cobra.Command{
-	Use: fmt.Sprintf("set-resource RESOURCE [--group GROUP] [--force] --mode <%s|%s|%s>",
-		api.Propagate, api.Remove, api.Ignore),
+	Use: fmt.Sprintf("set-resource RESOURCE [--group GROUP] [--force] --mode <%s|%s|%s|%s>",
+		api.Propagate, api.Remove, api.Ignore, api.AllowPropagate),
 	Short: "Sets the HNC configuration of a specific resource",
 	Example: fmt.Sprintf("  # Set configuration of a core type\n" +
 		"  kubectl hns config set-resource secrets --mode Ignore\n\n" +
@@ -40,7 +40,7 @@ var setResourceCmd = &cobra.Command{
 		flags := cmd.Flags()
 		group, _ := flags.GetString("group")
 		modeStr, _ := flags.GetString("mode")
-		mode := api.SynchronizationMode(cases.Title(language.English).String(modeStr))
+		mode := normalizeMode(modeStr)
 		force, _ := flags.GetBool("force")
 		config := client.getHNCConfig()
 
@@ -48,8 +48,8 @@ var setResourceCmd = &cobra.Command{
 		for i := 0; i < len(config.Spec.Resources); i++ {
 			r := &config.Spec.Resources[i]
 			if r.Group == group && r.Resource == resource {
-				if r.Mode == api.Ignore && mode == api.Propagate && !force {
-					fmt.Printf("Switching directly from 'Ignore' to 'Propagate' mode could cause existing %q objects in "+
+				if r.Mode == api.Ignore && (mode == api.Propagate || mode == api.AllowPropagate) && !force {
+					fmt.Printf("Switching directly from 'Ignore' to 'Propagate' mode or 'AllowPropagate' mode could cause existing %q objects in "+
 						"child namespaces to be overwritten by objects from ancestor namespaces.\n", resource)
 					fmt.Println("If you are sure you want to proceed with this operation, use the '--force' flag.")
 					fmt.Println("If you are not sure and would like to see what source objects would be overwritten," +
@@ -78,7 +78,19 @@ var setResourceCmd = &cobra.Command{
 
 func newSetResourceCmd() *cobra.Command {
 	setResourceCmd.Flags().String("group", "", "The group of the resource; may be omitted for core resources (or explicitly set to the empty string)")
-	setResourceCmd.Flags().String("mode", "", "The synchronization mode: one of Propagate, Remove or Ignore")
-	setResourceCmd.Flags().BoolP("force", "f", false, "Allow the synchronization mode to be changed directly from Ignore to Propagate despite the dangers of doing so")
+	setResourceCmd.Flags().String("mode", "", "The synchronization mode: one of Propagate, Remove, Ignore and AllowPropagate")
+	setResourceCmd.Flags().BoolP("force", "f", false, "Allow the synchronization mode to be changed directly from Ignore to Propagate or AllowPropagate despite the dangers of doing so")
 	return setResourceCmd
+}
+
+// normalizeMode takes a user-input mode and returns
+// a SynchronizationMode in a format HNC expects
+func normalizeMode(modeStr string) api.SynchronizationMode {
+	mode := api.SynchronizationMode(cases.Title(language.English).String(modeStr))
+
+	if mode == "Allowpropagate" {
+		return api.AllowPropagate
+	}
+
+	return mode
 }
--- a/internal/objects/reconciler.go
+++ b/internal/objects/reconciler.go
@@ -165,7 +165,7 @@ func (r *Reconciler) GetMode() api.SynchronizationMode {
 // treated as api.Ignore.
 func GetValidateMode(mode api.SynchronizationMode, log logr.Logger) api.SynchronizationMode {
 	switch mode {
-	case api.Propagate, api.Ignore, api.Remove:
+	case api.Propagate, api.Ignore, api.Remove, api.AllowPropagate:
 		return mode
 	case "":
 		log.Info("Sync mode is unset; using default 'Propagate'")
@@ -196,6 +196,11 @@ func (r *Reconciler) SetMode(ctx context.Context, log logr.Logger, mode api.Sync
 		}
 	}
 	return nil
+}
+
+// CanPropagate returns true if Propagate mode or AllowPropagate mode is set
+func (r *Reconciler) CanPropagate() bool {
+	return (r.GetMode() == api.Propagate || r.GetMode() == api.AllowPropagate)
 }
 
 // GetNumPropagatedObjects returns the number of propagated objects of the GVK handled by this object reconciler.
@@ -388,11 +393,13 @@ func (r *Reconciler) shouldSyncAsPropagated(log logr.Logger, inst *unstructured.
 	}
 
 	// If there's a conflicting source in the ancestors (excluding itself) and the
-	// the type has 'Propagate' mode, the object will be overwritten.
+	// the type has 'Propagate' mode or 'AllowPropagate' mode, the object will be overwritten.
 	mode := r.Forest.GetTypeSyncer(r.GVK).GetMode()
-	if mode == api.Propagate && srcInst != nil {
-		log.Info("Conflicting object found in ancestors namespace; will overwrite this object", "conflictingAncestor", srcInst.GetNamespace())
-		return true, srcInst
+	if mode == api.Propagate || mode == api.AllowPropagate {
+		if srcInst != nil {
+			log.Info("Conflicting object found in ancestors namespace; will overwrite this object", "conflictingAncestor", srcInst.GetNamespace())
+			return true, srcInst
+		}
 	}
 
 	return false, nil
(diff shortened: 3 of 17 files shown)


internal/selectors/selectors.go
```diff
@@ -158,6 +187,27 @@ func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
...
+	}
+	allSelector, err := strconv.ParseBool(allSelectorStr)
+	if err != nil {
+		// Returning false here in accordance to the Go standard
+		return false,
+			fmt.Errorf("invalid %s value: %w",
+				api.AnnotationAllSelector,
+				err,
+			)
+
```
adrianludwin, line 206: This is short enough that in HNC we'd typically put this all on one line:

```
return false, fmt.Errorf("invalid %s value: %w", api.AnnotationAllSelector, err)
```

test/e2e/issues_test.go
```diff
@@ -339,6 +339,58 @@ spec:
...
+
+	It("Should allow objects to be propagated only when a selector is set in Allow mode - issue #16", func() {
+		// Create a simple structure and get an object propagated
+		CreateNamespace(nsParent)
+		CreateNamespace(nsChild)
+		MustRun("kubectl hns set", nsChild, "--parent", nsParent)
+
+		MustRun("kubectl -n", nsParent, "create secret generic my-creds --from-literal=password=iamteamb")
+		// wait 2 seconds to give time for secret to propogate if it was to
+		time.Sleep(2 * time.Second)
```
adrianludwin, line 351: These sleeps _shouldn't_ be necessary - are they? `RunShouldNotContain` should repeatedly retry until it passes (up to a maximum of 2s).

test/e2e/issues_test.go
```diff
@@ -339,6 +339,58 @@ spec:
...
+		// secret does not show up in child namespae because we haven’t configured HNC to propagate secrets in HNCConfiguration.
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+		// after setting Allow mode, the object does not get propogated
+		MustRun("kubectl hns config set-resource secrets --mode Allow --force")
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
```
adrianludwin, line 356: There's no `Consistently` in the test package, so I'd add a 2s sleep before this line to give HNC time to do the wrong thing if it was going to.

adrianludwin (commented):
This looks great, just a bunch of mainly stylistic changes. Thanks for this!

/cc @rjbez17 
/ok-to-test

internal/hierarchyconfig/validator.go
```diff
@@ -246,10 +246,10 @@ func (v *Validator) getConflictingObjects(newParent, ns *forest.Namespace) []str
 	if newParent == nil {
 		return nil
 	}
-	// Traverse all the types with 'Propagate' mode to find any conflicts.
+	// Traverse all the types with 'Propagate' mode or 'Allow' mode to find any conflicts.
 	conflicts := []string{}
 	for _, t := range v.Forest.GetTypeSyncers() {
-		if t.GetMode() == api.Propagate {
+		if t.GetMode() == api.Propagate || t.GetMode() == api.Allow {
```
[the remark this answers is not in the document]
adrianludwin: (Here and in all other places where we explicitly check both modes)

internal/objects/reconciler.go
```diff
@@ -726,6 +728,16 @@ func (r *Reconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.U
 	case r.Mode == api.Remove:
 		return false
 
+	case r.Mode == api.Allow:
```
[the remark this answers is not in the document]
mzeevi: I initially thought about implementing it in this way, but was unsure on what parameters to pass from the webhooks since `Validator` has no way of knowing the propagation mode? So what should I pass for the extra parameter from the webhook functions (for example `func (v *Validator) getConflictingObjectsOfType`)?

mzeevi:
Thanks for the notes, Adrian! I will address everything in a new commit.

internal/objects/reconciler.go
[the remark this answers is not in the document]
mzeevi: What do you think? @adrianludwin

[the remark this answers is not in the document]
adrianludwin: The validator does have access to the propagation mode [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/347c5a3f92463d1157650172524d7743bdc6faa7/internal/objects/validator.go#L109) - is this what you mean?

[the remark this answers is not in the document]
mzeevi: Might have missed it, I'll use the same method on the other validators, thanks

mzeevi:
Should be fine now

internal/objects/validator.go
```diff
@@ -347,7 +376,10 @@ func (v *Validator) hasConflict(inst *unstructured.Unstructured) (bool, []string
 			// If the user have chosen not to propagate the object to this descendant,
 			// there shouldn't be any conflict reported here
 			nsLabels := v.Forest.Get(inst.GetNamespace()).GetLabels()
-			if ok, _ := selectors.ShouldPropagate(inst, nsLabels); ok {
+			fmt.Println("fuck1")
```
adrianludwin, line 379: Nope 🤣

internal/objects/validator.go
```diff
@@ -347,7 +376,10 @@ func (v *Validator) hasConflict(inst *unstructured.Unstructured) (bool, []string
...
+			propogateByDefault := v.isPropagateType(metav1.GroupVersionKind(gvk))
+			fmt.Println("fuck2")
+			if ok, _ := selectors.ShouldPropagate(inst, nsLabels, propogateByDefault); ok {
```
adrianludwin, line 382: Maybe just allow selectors.ShouldPropagate take the mode, so you can encapsulate the logic in there? It'll make the calling code shorter and nicer IMO.

internal/objects/validator_test.go
```diff
@@ -540,6 +540,34 @@ func TestUserChanges(t *testing.T) {
...
+	}, {
+		name: "Allow creation of object with valid all annotation",
+		fail: false,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationNoneSelector: "true",
```
adrianludwin, line 566: Should be `api.AnnotationAllSelector`?

internal/objects/validator_test.go
```diff
@@ -752,8 +780,13 @@ func TestCreatingConflictSource(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			// Setup
+			r := &Reconciler{
```
adrianludwin, line 783: Maybe say why you're adding this - e.g. because the validator needs to know the propagation mode.

Do you know why this ever worked before?

internal/selectors/selectors.go
```diff
@@ -15,7 +15,38 @@ import (
...
+	if !propagateByDefault {
+		// if propagateByDefault is false and there are no selectors - do not propagate,
+		// otherwise handle propagation as if propagateByDefault mode is true
+		if exists, err := SelectorExists(inst, nsLabels); err != nil {
+			return false, err
+		} else if !exists {
+			return false, nil
+		}
+	}
+
```
adrianludwin, line 49: I think this can just be moved to the last line of the function. Instead of just `return true, nil` it should be `return propagateByDefault, nil`.

Although if we change the function signature to accept the propagation mode, it should be `return mode == api.Propagate, nil` with a comment about what the alternatives are to 'propagate'.

test/e2e/issues_test.go
```diff
@@ -339,6 +339,57 @@ spec:
...
+		CreateNamespace(nsChild)
+		MustRun("kubectl hns set", nsChild, "--parent", nsParent)
+
+		MustRun("kubectl -n", nsParent, "create secret generic my-creds --from-literal=password=iamteamb")
+		// secret does not show up in child namespae because we haven’t configured HNC to propagate secrets in HNCConfiguration.
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+		// after setting AllowPropagate mode, the object does not get propogated
+		MustRun("kubectl hns config set-resource secrets --mode AllowPropagate --force")
+		time.Sleep(2 * time.Second)
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
```
adrianludwin, line 355: An even better test might be to first change the mode to `Propagate`, verify that the secret is created, and then change it to `AllowPropagate` and verify that it's _removed_ before going on with the rest of the test.

internal/objects/validator.go
mzeevi: Ooops I definitely thought I cleaned that up hahah

internal/kubectl/configset.go
```diff
@@ -40,16 +40,16 @@ var setResourceCmd = &cobra.Command{
 		flags := cmd.Flags()
 		group, _ := flags.GetString("group")
 		modeStr, _ := flags.GetString("mode")
-		mode := api.SynchronizationMode(cases.Title(language.English).String(modeStr))
```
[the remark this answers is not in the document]
mzeevi: Sounds good, I'll fix that

internal/objects/reconciler.go
```diff
@@ -463,7 +470,7 @@ func (r *Reconciler) syncPropagated(inst, srcInst *unstructured.Unstructured) (s
 func (r *Reconciler) syncSource(log logr.Logger, src *unstructured.Unstructured) {
 	// Update or create a copy of the source object in the forest. We now store
 	// all the source objects in the forests no matter if the mode is 'Propagate'
-	// or not, because HNCConfig webhook will also check the non-'Propagate' mode
+	// or not, because HNCConfig webhook will also check the non-'Propagate' or 'Allow' modes
```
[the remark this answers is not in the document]
mzeevi: Good catch, thought I caught all of them huh

internal/objects/validator_test.go
mzeevi: Yep

internal/objects/validator_test.go
mzeevi: I believe it worked previously because before there was only one mode that allowed propagation, so if it it was [checking for conflicts](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/347c5a3f92463d1157650172524d7743bdc6faa7/internal/objects/validator.go#L329-L357), it could assume it's `Propagate` mode since any other mode would have [already returned](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/347c5a3f92463d1157650172524d7743bdc6faa7/internal/objects/validator.go#L85-L86). Now that we've added another mode that allows propagation under certain conditions then we need to differentiate and therefore set up a reconcile with a specific mode

internal/selectors/selectors.go
mzeevi: Yes this is a lot cleaner and clearer, I agree.

mzeevi:
Made some changes as per your comments, @adrianludwin 

Thanks again!

internal/selectors/selectors.go
```diff
@@ -15,23 +15,39 @@ import (
...
+// based on the SynchronizationMode and on the selectors of the object.
+// Propagation will be done only with 'Propagate' or 'AllowPropagate' modes.
+// When selectors are set, 'AllowPropagate' follows the same logic as 'Propagate' mode,
+// when they are not set, then if 'Propagate' mode is used then propagate by default;
+// if 'AllowPropagate' mode is used then do not propagate by default.
+func ShouldPropagate(inst *unstructured.Unstructured, nsLabels labels.Set, mode api.SynchronizationMode) (bool, error) {
+	sel, err := GetSelector(inst)
+	if err != nil {
 		return false, err
 	} else if sel != nil && !sel.Matches(nsLabels) {
```
adrianludwin, line 28: No need for `else` here since `sel` is now defined outside of the if statement. Alternatively, maybe something like this would be easier to read:

```
propIfNotExcluded := (mode == api.Propagate)

if sel, err := GetSelector(inst); err != nil {
  return false, err
} else if sel != nil {
  propIfNotExcluded = true
  if !sel.Matches(nsLabels) {
    return false, nil
  }
}
```

Then at the end of the function, it's just `return propIfNotExcluded`. Up to you if you think that's clearer though, I know those nested `if` statements are a pain.

test/e2e/issues_test.go
```diff
@@ -339,6 +339,65 @@ spec:
...
+	It("Should allow objects to be propagated only when a selector is set in AllowPropagate mode - issue #16", func() {
+		// Create a simple structure and get an object propagated
+		CreateNamespace(nsParent)
+		CreateNamespace(nsChild)
+		MustRun("kubectl hns set", nsChild, "--parent", nsParent)
+
+		MustRun("kubectl -n", nsParent, "create secret generic my-creds --from-literal=password=iamteamb")
+		// secret does not show up in child namespae because we haven’t configured HNC to propagate secrets in HNCConfiguration.
+		time.Sleep(2 * time.Second)
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
```
adrianludwin, line 352: I'd take this out, this is tested elsewhere.

test/e2e/issues_test.go
```diff
@@ -339,6 +339,65 @@ spec:
...
+		// after setting Propagate mode, the object shows up in the child namespace
+		MustRun("kubectl hns config set-resource secrets --mode Propagate --force")
+		time.Sleep(2 * time.Second)
```
adrianludwin, line 355: Not needed, `RunShouldContain` repeats for 2s. All remaining `Sleeps` can and should be removed as long as the preceding command changes the state (e.g. from propagated to unpropagated or vice versa), which I believe is true. You only need to sleep if you're confirming that _nothing_ has changed, which we generally avoid doing.

test/e2e/issues_test.go
```diff
@@ -339,6 +339,65 @@ spec:
...
+		RunShouldContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+
+		// after setting AllowPropagate mode, the object is removed from the child namespace
+		MustRun("kubectl hns config set-resource secrets --mode AllowPropagate --force")
+		time.Sleep(2 * time.Second)
```
adrianludwin, line 360: Delete this and about 5-10 other instances below.

adrianludwin:
lgtm after these last changes are made. Ryan, please feel free to approve (or else I'll do it).

/assign @rjbez17

internal/selectors/selectors.go
mzeevi: Think it's clearer yes, need to check if `!sel.Empty()` as well.

test/e2e/issues_test.go
mzeevi: Understood, thanks

mzeevi:
Made some changes in accordance to the latest comments

test/e2e/issues_test.go
```diff
@@ -339,6 +339,54 @@ spec:
...
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+
+		// after adding a allSelector, the object shows up in the child namespace
+		MustRun("kubectl annotate secret my-creds -n", nsParent, "propagate.hnc.x-k8s.io/all=true")
+		RunShouldContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+		// after removing a allSelector, the object does not show up in the child namespace
+		MustRun("kubectl annotate secret my-creds -n", nsParent, "propagate.hnc.x-k8s.io/all-")
+		RunShouldNotContain("my-creds", defTimeout, "kubectl -n", nsChild, "get secrets")
+
+		MustRun("kubectl hns config set-resource secrets --mode Ignore")
```
adrianludwin, line 388: Sorry I didn't notice this until now - if the test fails, this line will never run. Can you add it to the BeforeEach() function on line 23 of this file, and mention this test as the reason why this needs to be reset? That's my last comment, promise :)

mzeevi: No problem, and all comments are welcome :)

mzeevi:
Should be fine now @adrianludwin

adrianludwin (commented):
/lgtm
/approve

Thanks for this change!
I'd appreciate it if you could also submit a docs change afterwards. Here are the sections that should be updated:
* [How-to: selectors](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#use-limit-propagation)
* [How-to: admin resources](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#admin-resources)
* [Concepts: exceptions](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#basic-exceptions)

Please indicate that these features will be beta in v1.1. Thanks!

Issue #221, opened 2022-08-29
Add documentation for inclusive propagation


adrianludwin:
/assign @mzeevi 

See https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/217#pullrequestreview-1088726641

adrianludwin:
@mzeevi I can't assign this to you formally but here it is FYI :)

mzeevi:
Addressed this at #222, let me know what you think. @adrianludwin

adrianludwin:
/remove-lifecycle rotten

This was actually fixed properly.

PR #222, opened 2022-08-30
Doc for opt-in propagation

mzeevi:
Addresses issue #221, documentation is added on the `AllowPropagate`
sync mode and on the `all` selector.

Signed-off-by: mzeevi <meytar80@gmail.com>

### DIFF
--- a/docs/user-guide/concepts.md
+++ b/docs/user-guide/concepts.md
@@ -319,10 +319,10 @@ application.
 for policy purposes.*
 
 <a name="basic-exceptions"/>
-
+ 
 ### Exceptions and propagation control
 
-By default, HNC propagates _all_ objects of a [specified type](how-to.md#admin-resources)
+HNC typically propagates _all_ objects of a [specified type](how-to.md#admin-resources)
 from ancestor namespaces to descendant namespaces. However, sometimes this is
 too restrictive, and you need to create ***exceptions*** to certain policies. For example:
 
@@ -351,6 +351,20 @@ from modifying the object, and will identify the objects that would have been
 overwritten by your actions. You can then rewrite the exception to safely
 exclude those objects, or else delete the conflicting objects to allow them to
 be replaced.
+
+#### (Beta in v1.1) Opt-in propagation
+The `Propagate` mode propagates all objects unless directed to otherwise via a selector,
+using exceptions. By contrast, opt-in propagation, as set by the `AllowPropagate` 
+mode, doesn't propagate objects unless directed to by a selector. That is, for an object 
+with a selector, its behaviour will be identical in both the `Propagate` and `AllowPropagate` modes; 
+the only difference in behaviours is for objects without a selector.
+
+For example:
+ 
+* A Secret exists on a namespace but we don't want this secret to be propagated
+to all subnamespaces by default but instead only to one namespace of our choosing. 
+So you can choose to propagate to that _one_ child namespace using `AllowPropagate`
+and exceptions.
 
 #### Built-in exceptions
 
--- a/docs/user-guide/how-to.md
+++ b/docs/user-guide/how-to.md
@@ -366,7 +366,17 @@ can use any of the following annotations:
 
 * **`propagate.hnc.x-k8s.io/none`**: Setting `none` to `true` (case insensitive)
   will result in the object not propagating to _any_ descendant namespace. Any
-  other value will be rejected.
+  other value will be rejected. This is only useful in `Propagate`
+  [synchronization mode](#modify-the-resources-propagated-by-hnc).
+
+* **`propagate.hnc.x-k8s.io/all`**: Setting `all` to `true` (case insensitive)
+  will result in the object propagating to _any_ descendant namespace. Any
+  other value will be rejected. This is only useful in `AllowPropagate`
+  [synchronization mode](#modify-the-resources-propagated-by-hnc).
+
+Warning: only use one of the propagation annotations on any one object at a time. 
+Interactions between multiple propagation annotations is undefined and may change 
+from release to release.
 
 For example, consider a case with a parent namespace with three child
 namespaces, and the parent namespace has a secret called `my-secret`. To set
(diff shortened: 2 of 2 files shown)


docs/user-guide/concepts.md
```diff
@@ -322,28 +322,46 @@ for policy purposes.*
...
+[synchronization mode](how-to.md#modify-the-resources-propagated-by-hnc), or if HNC only
+propagates objects of a specified type when selectors are used, using the `AllowPropagate`
+synchronization mode.
+
+When `Propagate` [synchronization mode](how-to.md#modify-the-resources-propagated-by-hnc) is
+configured for a [specified type](how-to.md#admin-resources), HNC propagates _all_ objects of this type
+from ancestor namespaces to descendant namespaces. When `AllowPropagate` synchronization mode
+is configured for a specified type, HNC does not propagates _any_ objects of this type
+from ancestor namespaces to descendant namespaces. ***Exceptions*** are used to create
+certain policies to modify the behavior of the propagation. For example:
```
adrianludwin, line 336: This text goes live immediately so we don't want people using HNC v1.0 to think that this is available when it's not. Also, for at least one minor release, we'll call this "beta."

So how about this:
* Restore the original text, but add a new section (above the one called "Built-in exceptions") called "[Beta in v1.1] Inclusive propagation" and describe `AllowPropagate` there.
* Soften the intro line from "By default, HNC propagates _all_ objects" to "HNC typically propagates _all_ objects" which is becoming more and more accurate.

docs/user-guide/how-to.md
```diff
@@ -366,7 +366,15 @@ can use any of the following annotations:
 
 * **`propagate.hnc.x-k8s.io/none`**: Setting `none` to `true` (case insensitive)
   will result in the object not propagating to _any_ descendant namespace. Any
-  other value will be rejected.
+  other value will be rejected. The behavior would be the same as setting
+  [synchronization mode](#modify-the-resources-propagated-by-hnc) to 
+  `AllowPropagate` with no selectors at all.
```
adrianludwin, line 371: I think this is a bit confusing - maybe better to say something like "This is only useful in the `Propagate` mode." Ditto in the next section, I might just say "This is only useful in the `AllowPropagate` mode."

docs/user-guide/how-to.md
```diff
@@ -388,6 +396,12 @@ To set `my-secret` not to propagate to any namespace, you can use:
 kubectl annotate secret my-secret -n parent propagate.hnc.x-k8s.io/none=true
 ```
 
+To set `my-secret` to propagate to any namespace, you can use:
```
adrianludwin, line 399: Suggestion: To set `my-secret` to propagate to any namespace when the mode is `AllowPropagate`, you can use:

docs/user-guide/how-to.md
```diff
@@ -677,6 +691,10 @@ The most important type of configuration is the way each object type
 
 * **Propagate:** propagates objects from ancestors to descendants and deletes
   obsolete descendants.
+* **AllowPropagate:** inclusive propagation - only propagates objects from ancestors 
+  to descendants and deletes obsolete descendants when
+  [selectors](#limit-the-propagation-of-an-object-to-descendant-namespaces) are set
+  on the object.
```
adrianludwin, line 697: I might be a bit more precise here: "when at least one selector is set on the object."

docs/user-guide/how-to.md
```diff
@@ -366,7 +366,15 @@ can use any of the following annotations:
...
-  other value will be rejected.
+  other value will be rejected. The behavior would be the same as setting
+  [synchronization mode](#modify-the-resources-propagated-by-hnc) to 
+  `AllowPropagate` with no selectors at all.
+
+* **`propagate.hnc.x-k8s.io/all`**: Setting `all` to `true` (case insensitive)
+  will result in the object propagating to _any_ descendant namespace. Any
+  other value will be rejected. The behavior would be the same as setting
+  [synchronization mode](#modify-the-resources-propagated-by-hnc) to 
+  `Propagate` with no selectors at all.
```
adrianludwin, line 377: I think we should add a new paragraph below this list saying something like: "Warning: only use one of the propagation annotations on any one object at a time. Interactions between multiple propagation annotations is undefined and may change from release to release."

adrianludwin (commented):
Thanks! Just a few requested changes.

docs/user-guide/how-to.md
mzeevi: Yes, I was wondering how to phrase it, this sounds better

docs/user-guide/how-to.md
mzeevi: Interactions between annotations are checked in the [integ testing ](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/internal/objects/reconciler_test.go#L42-L176)

You think the behavior might change in the future so a warning would be better?

docs/user-guide/how-to.md
mzeevi: Will change it

docs/user-guide/how-to.md
adrianludwin: Oh, I forgot about that. If it's straightforward to document, then maybe you can document it? Otherwise we can tell people not to rely on it :)

docs/user-guide/concepts.md
```diff
@@ -352,6 +352,19 @@ overwritten by your actions. You can then rewrite the exception to safely
 exclude those objects, or else delete the conflicting objects to allow them to
 be replaced.
 
+#### (Beta in v1.1) Inclusive propagation
+Inclusive propagation allows you to choose which objects to propagate by using exceptions.
```
adrianludwin, line 356: This first sentence doesn't really show how inclusive propagation is different from the regular sort. How about this: instead of "inclusive propagation," call it "Opt-in propagation." Then the first sentence changes to:

The `Propagate` mode propagates all objects unless directed to otherwise via a selector. By contrast, opt-in propagation, as set by the `AllowPropagate` mode, doesn't propagate objects _unless_ directed to by a selector. That is, for an object with a selector, its behaviour will be identical in both the `Propagate` and `AllowPropagate` modes; the only difference in behaviours is for objects without a selector.

docs/user-guide/how-to.md
```diff
@@ -737,20 +754,20 @@ spec:
...
-Adding a new resource in the `Propagate` mode is potentially dangerous, since
-there could be existing objects of that resource type that would be overwritten
-by objects of the same name from ancestor namespaces. As a result, the HNS
-plugin will not allow you to add a new resource directly in the `Propagate`
-mode.  Instead, to do so safely:
+Adding a new resource in the `Propagate` mode or `AllowPropagate` mode is potentially 
+dangerous, since there could be existing objects of that resource type that 
+would be overwritten by objects of the same name from ancestor namespaces. 
+As a result, the HNS plugin will not allow you to add a new resource directly 
+in the `Propagate` mode or `AlowPropagate`.  Instead, to do so safely:
```
adrianludwin, line 761: Typo: `AllowPropagate`

adrianludwin (commented):
lgtm apart from two changes (including that we should change the name of the feature). Can you please squash the commits as well when you next push? Thanks!

docs/user-guide/how-to.md
mzeevi: I've added a warning for now, might doc it in a future PR though

mzeevi:
Should be OK now

adrianludwin (commented):
/lgtm
/approve

Issue #228, opened 2022-09-23
Resources deleted when HNC restarts despite existing previously 


Alan01252:
When HNC restarts the propagated objects can be deleted unnecessarily if the parent namespace hasn't yet fully synced.
This is problematic especially when propagating network rules as can lead to loss of service whilst the network rules are re-created.

Digging through the code the problem appears to be that the ActivitiesHalted flag is removed too early.
It's removed when the parent namespace is added to the forest, not when all the objects inside the parent namespace have been synced.

I'm been trying to think of ways to fix this in the code but have come up short.

Hoping someone here can please advise!

version: 1.0.0

Alan01252:
This is an example log line:

`{"level":"info","ts":1663928171.33828,"logger":"hierarchyconfig.reconcile","msg":"Setting ConditionActivitiesHalted: parent doesn't exist (or hasn't been synced yet)","rid":11,"ns":"acp-shrd-srv-service-auth","parent":"acp-shrd"}
{"level":"info","ts":1663928171.3384438,"logger":"hierarchyconfig.reconcile","msg":"New namespace found","rid":11,"ns":"acp-shrd-srv-service-auth"}
{"level":"info","ts":1663928171.3384647,"logger":"hierarchyconfig.reconcile","msg":"Namespace's managed and/or tree labels have been updated","rid":11,"ns":"acp-shrd-srv-service-auth"}
{"level":"info","ts":1663928171.3384745,"logger":"hierarchyconfig.reconcile","msg":"Setting ActivitiesHalted on namespace","rid":11,"ns":"acp-shrd-srv-service-auth","conditions":[{"type":"ActivitiesHalted","status":"True","lastTransitionTime":"1970-01-01T00:00:00Z","reason":"ParentMissing","message":"Parent \"acp-shrd\" does not exist"}]}
{"level":"info","ts":1663928173.433423,"logger":"RoleBinding.reconcile","msg":"Namespace has 'ActivitiesHalted' condition; will not touch propagated object","rid":37,"trigger":"acp-shrd-srv-service-auth/EKS-Shared-Admin","haltedRoot":"acp-shrd-srv-service-auth","suppressedAction":"remove"}
[...]

PR #229, opened 2022-09-26
Fix resources not managed by HNC being deleted

Alan01252:
This commit fixes the problem that when HNC restarts the source objects are deleted and then re-created.

When a source object is deleted the code now checks the decedents for existing source objects and adds an annotation to them to make sure that only sources hnc is aware of are deleted.

It also makes sure that all source files are synchronised with the forest even if they are just propergated objects so that the forest is more accurate reflection of what's on the cluster.

### DIFF
--- a/internal/forest/namespaceobjects.go
+++ b/internal/forest/namespaceobjects.go
@@ -1,6 +1,7 @@
 package forest
 
 import (
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -28,11 +29,25 @@ func (ns *Namespace) HasSourceObject(gvk schema.GroupVersionKind, oo string) boo
 }
 
 // DeleteSourceObject deletes a source object by name.
-func (ns *Namespace) DeleteSourceObject(gvk schema.GroupVersionKind, nm string) {
+func (ns *Namespace) DeleteSourceObject(log logr.Logger, gvk schema.GroupVersionKind, nm string) {
+
+	log.Info("Deleting source object in namespace", "namespace", ns.Name(), "gvk", gvk, "nm", nm)
 	delete(ns.sourceObjects[gvk], nm)
 	// Garbage collection
 	if len(ns.sourceObjects[gvk]) == 0 {
 		delete(ns.sourceObjects, gvk)
+	}
+
+	a := make(map[string]string)
+	a["hnc.x-k8s.io/delete"] = "true"
+	for _, dec := range ns.DescendantNames() {
+		sObj := ns.forest.Get(dec).GetSourceObject(gvk, nm)
+		if sObj != nil {
+			log.Info("Found object adding deletion annotation", "namespace", ns.Name(), "gvk", gvk, "nm", nm, "dec", dec)
+			sObj.SetAnnotations(a)
+			continue
+		}
+		log.Info("Source object not found", "namespace", ns.Name(), "gvk", gvk, "nm", nm, "dec", dec)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


Alan01252:
As stated on the issue https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/228 I do not expect this to be merged. I have opened just to spark some conversation around the approach.

internal/objects/validator.go
```diff
@@ -165,7 +165,7 @@ func (v *Validator) handle(ctx context.Context, req *request) admission.Response
```
Alan01252, line 165: I'm not sure about this at all. 

The check was to prevent desendents being overwritten, *but* once the above fix is in place to add source objects for descendants this validation check fails even if you're just updating the parent object. I flipped it to ancestors just so flux stopped getting dry-run failures in our environment.

I'm not sure the check makes much sense if validating against the forest because if the forest contains source objects, etc., already, you've got to assume HNC created them and any conflicts should have caused errors earlier in the process.

internal/forest/namespaceobjects.go
```diff
@@ -28,12 +29,26 @@ func (ns *Namespace) HasSourceObject(gvk schema.GroupVersionKind, oo string) boo
...
-func (ns *Namespace) DeleteSourceObject(gvk schema.GroupVersionKind, nm string) {
+func (ns *Namespace) DeleteSourceObject(log logr.Logger, gvk schema.GroupVersionKind, nm string) {
+
+	log.Info("Deleting source object in namespace", "namespace", ns.Name(), "gvk", gvk, "nm", nm)
 	delete(ns.sourceObjects[gvk], nm)
 	// Garbage collection
 	if len(ns.sourceObjects[gvk]) == 0 {
 		delete(ns.sourceObjects, gvk)
 	}
+
```
Alan01252, line 40: add an annotation which prevents any resource hnc was not aware of being deleted.

internal/objects/reconciler.go
```diff
@@ -411,34 +412,55 @@ func (r *Reconciler) getTopSourceToPropagate(log logr.Logger, inst *unstructured
...
+	ns.DeleteSourceObject(log, r.GVK, inst.GetName())
 	stats.OverwriteObject(r.GVK)
 
 	// If no source object exists, delete this object. This can happen when the source was deleted by
 	// users or the admin decided this type should no longer be propagated.
 	if srcInst == nil {
-		return actionRemove, nil
+		log.Info("Unable to find srcInst removing object", "namespace", ns.Name(), "namespaceParent", ns.Parent(), "inst", inst.GetName())
+
+		if srcObj == nil {
```
Alan01252, line 453: this does seem to get hit more frequently than I'd imagine, I think just because of cycles, but doesn't seem to cause any problems

internal/objects/reconciler.go
```diff
@@ -411,34 +412,55 @@ func (r *Reconciler) getTopSourceToPropagate(log logr.Logger, inst *unstructured
...
+			log.Info("src object not found for instance", "namespace", ns.Name(), "namespaceParent", ns.Parent(), "inst", inst.GetName(), "gvk", r.GVK)
+			return actionNop, nil
+		}
+
+		if srcObj.GetAnnotations()["hnc.x-k8s.io/delete"] == "true" {
```
Alan01252, line 458: Makes sure we're only deleting things we're aware of

internal/objects/reconciler.go
```diff
@@ -451,14 +473,17 @@ func (r *Reconciler) syncPropagated(inst, srcInst *unstructured.Unstructured) (s
...
+
+		log.Info("src object doesn't exist creating", "namespace", ns.Name(), "namespaceParent", ns.Parent(), "inst", inst.GetName())
 		return actionWrite, srcInst
 	}
 
+	log.Info("src object exists on api server but not in forest", "namespace", ns.Name(), "namespaceParent", ns.Parent(), "inst", inst.GetName())
 	// The object already exists and doesn't need to be updated. This will typically happen when HNC
 	// is restarted - all the propagated objects already exist on the apiserver. Record that it exists
 	// for our statistics.
+	r.syncSource(log, inst)
```
Alan01252, line 485: Makes sure the forest is up to date with everything that's been reconciled, especially important on restarts when the forest is new

Issue #230, opened 2022-09-29
Unable to delete Subnamespaceanchor with long name


gazpwc:
Hi,

somehow (maybe during a downtime of the hnc controller?) we managed to create a Subnamespaceanchor with a name having more than 63 characters. For obvious reasons, the controller is not able to create a corresponding namespace and the anchor remains in:

```
status:
  status: Missing
```

And unfortunately we are not even able to delete the invalid Subnamespaceanchor again.
```
kubectl delete subns very-very-....-long-name -n my-root-ns

The SubnamespaceAnchor "very-very-....-long-name" is invalid: metadata.name: Invalid value: "very-very-....-long-name": not a valid namespace name: must be no more than 63 characters
```

The HNC Controller logs this message:
```
{"level":"info","ts":1664440307.2302854,"logger":"anchor.validate","msg":"Denied","ns":"my-root-ns","nm":"very-very-....-long-name","op":"DELETE","user":"masterclient","code":422,"reason":"Invalid","message":"SubnamespaceAnchor.hnc.x-k8s.io \"very-very-....-long-name\" is invalid: metadata.name: Invalid value: \"very-very-....-long-name\": not a valid namespace name: must be no more than 63 characters"}
```

Is there any way of bypassing the validation and get rid of that invalid Subnamespaceanchor?

adrianludwin:
This is definitely a clear bug and we should fix it. Sorry! In the meantime, you'll have to temporarily delete the admission webhook config and then reinstall it after you've deleted the anchor.

gazpwc:
Thanks for the quick feedback. We will have a look into this to get rid of these objects.

After reading the docs (https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#uninstall-hnc-from-a-cluster) just one more question:

Just deleting the webhook by `kubectl delete validatingwebhookconfiguration.admissionregistration.k8s.io hnc-validating-webhook-configuration` will not cause any data/namespace/subnamespaceanchor/etc loss? So, after reinstalling it, everything is working as before?

adrianludwin:
That's correct, the worst thing that can happen when you delete a webhook
config is that K8s will allow you to make changes that it would otherwise
have blocked. So unless some other controller is going crazy trying to
delete things (which really, *really* shouldn't be happening) you'll be
fine.

If you want to be absolutely sure, you can back up your HNC objects as
described here:
[...]

PR #234, opened 2022-10-30
Fix validation of Subnamespaceanchor deletion

mochizuki875:
What type of PR is this?:
bug fix

What this PR does / why we need it:
Fix validation of Subnamespaceanchor deletion.
Skip char check at deletion event.

Tested: 
Added new unit test for my change; verified that the test passed.

Which issue(s) this PR fixes:
#230


mochizuki875:
@adrianludwin PTAL?

adrianludwin:
/lgtm
/ok-to-test
/approve

Sorry for the delay and thanks for this!

PR #237, opened 2022-11-13
Refactor some HRQ code for readability

adrianludwin:
I found some of this code very difficult to read, and there was a bunch of stuff in the 'utils' package that was being exported but not used (this is because the library was originally copied from K8s, but it's been heavily modified since then). Hopefully others find these cleanups useful to.

I did simplify the TryUseResources function to allow decreasing usage. This opens up a *very* slight race condition at the cost of making the function more robust to future changes in K8s and significantly easier to understand and test.

### DIFF
--- a/internal/forest/namespacehrq.go
+++ b/internal/forest/namespacehrq.go
@@ -36,18 +36,12 @@ type usage struct {
 	subtree v1.ResourceList
 }
 
-// TryUseResources checks resource limits in the namespace and its ancestors
-// when given proposed absolute (not delta) resource usages in the namespace. If
-// the proposed resource usages are the same as the current usages, of course
-// it's allowed and we do nothing. If there are any changes in the usages, we
-// only check to see if the proposed changing usages are allowed. If any of them
-// exceed resource limits, it returns an error; otherwise, it returns nil.
-// Callers of this method are responsible for updating resource usage status of
-// the HierarchicalResourceQuota objects.
-//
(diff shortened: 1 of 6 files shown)


adrianludwin:
/cc @mzeevi 

The next PR (which I have ready to go) adds a resync checker to make sure the incremental updates are always correct. That should fully complete the HRQ functionality.

internal/forest/namespacehrq.go
```diff
@@ -62,38 +56,21 @@ type usage struct {
...
-// When ResourceQuota reconciler receives the lock later, it will notice the
-// ResourceQuota.Status.Used is the same as in-memory usages and will not
-// update in-memory usages.
+// This may allow one weird case where a user may be allowed to use something they weren't supposed
+// to. Let's say you're well over your limit, and then in quick succession, some resources are
+// deleted, and some _fewer_ are added, but enough to still go over the limit. In that case, there's
+// a race condition between this function being called, and the RQ reconciler updating the baseline
+// resource usage. If this function wins, it will look like resource usage is decreasing, and will
+// be incorrectly allowed. If the RQ reconciler runs first, we'll see that the usage is incorrectly
+// _increasing_ and it will be disallowed. However, I think the simplicity of not trying to prevent
```
rjbez17, line 71: > I think the simplicity of not trying to prevent this (hopefully very unlikely) corner case is more valuable than trying to catch it.


Agreed.

PR #239, opened 2022-11-14
Expose port 8081 in the manager pods

### LINKED ISSUE #233: Missing port 8081 in Manager manifest
The manager pod is using port 8081 for its health probes. That port is, however, not exposed to the user in the manifest, which makes it harder than necessary to know that the port is indeed open. 

I suggest that the port is added to the list of exposed ports in the container.

If other containers have the same situation, the same fix should go for them.

https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/e9d8bb80e54ecfe735235ce608905e6dfbf757ca/config/manager/manager.yaml#L52-L58

lindhe:
Resolves #233

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -56,6 +56,9 @@ spec:
         - containerPort: 8080
           name: metrics
           protocol: TCP
+        - containerPort: 8081
+          name: healthz
+          protocol: TCP
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true


adrianludwin:
@lindhe do you think you'll be able to do all the CLA stuff? Otherwise lgtm, thanks!

/ok-to-test
/lgtm
/approve

lindhe:
1. Is `healthz` a good name? It's also used for `readyz`, apparently.
2. I've tried to locate other instances of this, but haven't been able to. Do you know of any other Deployment/Pod where this change should be made?

adrianludwin:
This is the only pod so it _should_ be the only place that needs this port exposed. Even in HA mode, we just make more copies of this one deployment ([here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/config/variants/ha-webhooks-cc/kustomization.yaml)) so this change should hopefully cover all cases.

Yes, healthz is fine, I _think_ it predated readyz and is the more "canonical" name.

PR #243, opened 2022-11-20
fix(concepts): fix the link for the resource quotas restriction example

sanposhiho:
I guess `#basic-exceptions` is a more appropriate section here.


adrianludwin (commented):
/lgtm
/approve
/ok-to-test

Issue #252, opened 2023-01-30
Unit tests could not run


mochizuki875:
## What happened?
Unit tests (`make test` as shown in [Development Workflow](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/e4ac07338ebf68144e81d5393fe3029d194e3474/docs/contributing.md#development-workflow)) could not run because the controlplane could not be started using testenv.

## What did you expect to happen?
I expect the controlplane to start using envtest and the unit tests are able to run.

## How can we reproduce it (as minimally and precisely as possible)?
Run unit tests by executing `make test`.
The error will occur as below.

```console
$ make test
...
[BeforeSuite] [FAILED] [10.006 seconds]
[BeforeSuite] 
/home/ubuntu-user/go/hierarchical-namespaces/internal/anchor/reconciler_test.go:20

  Begin Captured GinkgoWriter Output >>
    STEP: configuring test environment 01/26/23 13:21:34.371
    STEP: starting test environment 01/26/23 13:21:34.371
    1.6747069043722517e+09	DEBUG	controller-runtime.test-env	starting control plane
    1.674706904375007e+09	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 0, "error": "fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
    sigs.k8s.io/controller-runtime/pkg/envtest.(*Environment).startControlPlane
    	/home/ubuntu-user/go/hierarchical-namespaces/vendor/sigs.k8s.io/controller-runtime/pkg/envtest/server.go:330
    sigs.k8s.io/controller-runtime/pkg/envtest.(*Environment).Start
    	/home/ubuntu-user/go/hierarchical-namespaces/vendor/sigs.k8s.io/controller-runtime/pkg/envtest/server.go:260
    sigs.k8s.io/hierarchical-namespaces/internal/integtest.HNCBeforeSuite
    	/home/ubuntu-user/go/hierarchical-namespaces/internal/integtest/setup.go:82
    github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func2
    	/home/ubuntu-user/go/hierarchical-namespaces/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:605
    1.6747069043755567e+09	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 1, "error": "fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
[...]

PR #253, opened 2023-01-30
Fix Unit tests could not run

mochizuki875:
## What type of PR is this?:
/kind bug

Fix `Makefile` to setup envtest.
When `make test` is executed, download `envtest-setup` and envtest binaries locally(`/hack`) if necessary.
This fix references [Makefile generated by kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/2ea935fdbb28a34fc1bc705e2e3b3468d43761e9/docs/book/src/cronjob-tutorial/testdata/project/Makefile).

## Tested:
`make test` fails without this fix and passes with it.

## Which issue(s) this PR fixes:
#252

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.26.0
+
 # Adding .PHONY to hold command shortcuts.
 .PHONY: release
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -100,12 +103,20 @@ all: test docker-build
...
 
-test-only:
+LOCALBIN ?= $(shell pwd)/hack
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+test-only: envtest
 	@echo
 	@echo "If tests fail due to no matches for kind \"CustomResourceDefinition\" in version \"apiextensions.k8s.io/v1\","
 	@echo "please remove the old kubebuilder and reinstall it - https://book.kubebuilder.io/quick-start.html#installation"
 	@echo
```
adrianludwin, line 113: Can these lines be removed now? I think the makefile now does what this is telling the user to do?

adrianludwin (commented):
Very cool change, thanks! That's been an annoying part of the setup for a while now.

Just one question before we submit. If you make a change, please squash your commits so there's only one commit in this PR - thanks :)

/ok-to-test

Makefile
mochizuki875: @adrianludwin 
> Can these lines be removed now? 

I think so.  
Now, `kubebuilder` is installed as single binary like [this(`/usr/local/bin/kubebuilder`)](https://kubebuilder.io/quick-start.html#installation) not like [previously](https://github.com/kubernetes-sigs/controller-runtime/issues/90#issuecomment-494878527).  
So these lines may be useless.

adrianludwin:
Actually, there's a better way to do this - I've added setup-envtest to be hermetic as a part of #257. So you can cancel this PR, but thanks for submitting it as I've copied some of the key lines into my PR!

adrianludwin:
This should be fixed now, please lmk if it's not!

mochizuki875:
@adrianludwin 
This PR was fixed in #257 so I'll close this PR.
Thx!

Issue #246, opened 2023-01-11
Update conflict error occur


mochizuki875:
### What happened?
Resource update conflict error occurs when create/delete `subnamespace`.

### What did you expect to happen?
This conflict error dosen't occur.

### How can we reproduce it (as minimally and precisely as possible)?
Create a `subnamespace`.

```console
$ kubectl hns create subns-a -n test
Successfully created "subns-a" subnamespace anchor in "test" namespace
```

Check `hnc-controller-manager` log.
```console
$ kubeclt -n hnc-system logs hnc-controller-manager-xxxxxxxx
...
{"level":"error","ts":1673418845.0319858,"logger":"hierarchyconfig.reconcile","msg":"while updating apiserver","rid":183,"ns":"subns-a","error":"Operation cannot be fulfilled on namespaces \"subns-a\": the object has been modified; please apply your changes to the latest version and try again"}
{"level":"error","ts":1673418845.113606,"logger":"controller.hierarchyconfiguration","msg":"Reconciler error","reconciler group":"hnc.x-k8s.io","reconciler kind":"HierarchyConfiguration","name":"hierarchy","namespace":"subns-a","error":"Operation cannot be fulfilled on namespaces \"subns-a\": the object has been modified; please apply your changes to the latest version and try again"}
...
```

This seems to occur probabilistically with the creation and deletion of `subnamespaces`.

### Kubernetes version

```console
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-12T10:57:26Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-25T19:35:11Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"linux/amd64"}

$ kind version
kind v0.17.0 go1.19.2 linux/amd64
```

mochizuki875:
I think these conflicts are caused by two reasons.

[...]

PR #255, opened 2023-01-30
Replace Update() to Patch()

mochizuki875:
## What type of PR is this?:
/kind bug

Replaced `Update()` to `Patch()` in `anchor.reconcile` and `hierarchyconfig.Reconcile` to avoid resource update conflict error.

## Tested:
Ran unit tests and e2e tests.

## Which issue(s) this PR fixes:
#246

### DIFF
--- a/internal/anchor/reconciler.go
+++ b/internal/anchor/reconciler.go
@@ -67,7 +67,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Get instance from apiserver. If the instance doesn't exist, do nothing and early exist because
 	// HCR watches anchor and (if applicable) has already updated the parent HC when this anchor was
 	// purged.
-	inst, err := r.getInstance(ctx, pnm, nm)
+	baseInst, err := r.getInstance(ctx, pnm, nm)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			log.Info("Anchor has been deleted")
@@ -75,6 +75,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 		return ctrl.Result{}, err
 	}
+	inst := baseInst.DeepCopy()
 
 	// Anchors in unmanaged namespace should be ignored. Make sure it
 	// doesn't have any finalizers, otherwise, leave it alone.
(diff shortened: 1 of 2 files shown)


adrianludwin:
Thanks for this! A few questions:
* Do you know _why_ `patch` fixes this problem?
* Should this be applied to the other resource types? Or is it only affecting this one, and if so, why?

internal/anchor/reconciler.go
```diff
@@ -67,22 +67,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Get instance from apiserver. If the instance doesn't exist, do nothing and early exist because
 	// HCR watches anchor and (if applicable) has already updated the parent HC when this anchor was
 	// purged.
-	inst, err := r.getInstance(ctx, pnm, nm)
+	baseInst, err := r.getInstance(ctx, pnm, nm)
```
adrianludwin, line 70: Nit: I prefer `origInst` to focus on what's changed - I find "base" to be ambiguous here. To my eyes, "orig" makes it more clear that we'll be doing a diff on it later.

internal/anchor/reconciler.go
```diff
@@ -341,15 +342,20 @@ func (r *Reconciler) getInstance(ctx context.Context, pnm, nm string) (*api.Subn
...
 		if err := r.Create(ctx, inst); err != nil {
 			log.Error(err, "while creating on apiserver")
 			return err
 		}
 	} else {
-		if err := r.Update(ctx, inst); err != nil {
-			log.Error(err, "while updating on apiserver")
+		err := r.Patch(ctx, inst, client.MergeFrom(baseInst))
+		if err != nil {
+			if apierrors.IsNotFound(err) {
```
adrianludwin, line 354: Was this needed for functionality or just for those (very rare) cases where it makes the logs messy?

internal/hierarchyconfig/reconciler.go
```diff
@@ -137,8 +140,14 @@ func (r *Reconciler) handleUnmanaged(ctx context.Context, log logr.Logger, nm st
...
-		if err := r.Update(ctx, inst); err != nil {
-			log.Error(err, "while removing finalizers on unmanaged namespace")
+
+		err := r.Patch(ctx, inst, client.MergeFrom(baseInst))
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Info("Singleton has been deleted")
+				return nil
+			}
+			log.Error(err, "while removing finalizers and children on unmanaged singleton")
```
adrianludwin, line 150: I wonder if this could be factored out, e.g.:

```
err := r.Patch(...)
if err != nil {
  return somelib.handleMaybeDeletedError(err, "Singleton", "while removing finalizers and children...")
}
```

wdyt? That would make the code cleaner throughout.

adrianludwin:
Closing as described in #246 

/close

PR #258, opened 2023-02-12
build(arm64): publish built binaries for linux/arm64

pickledcoder:
This is to publish the [already built binaries ](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/Makefile#L125)for `linux/arm64`.
It is needed when running with Docker on Apple M1.


adrianludwin (commented):
/ok-to-test
/lgtm
/approve

PR #257, opened 2023-02-10
Update to Go 1.20 and latest KRTE for e2e tests

adrianludwin:
E2Es broke a few days ago, likely when test-infra upgraded to 1.20. I
also couldn't run staticcheck on my Google workstation, which has 1.20,
with the same failure message. Upgrading staticcheck fixed the problem;
I also took the opportunity to match the test-infra krte image as well
as the Golang version (which caused some minor reformatting), and also
made the integ tests hermetic by including the setup-envtest tool (bug
#252, based on an approach suggested in #253).

Tested: couldn't build on my workstation on Go 1.20 before this change;
can after it. Can also run integ tests after deleting local output of
setup-envtest.

Fixes #252

PR #261, opened 2023-02-24
Update Dockerfile to Go 1.20

mochizuki875:
In #257, Go version has been updated `1.20` from `1.17`.
However the Go version of `Dockerfile` was still `1.17`, so I fixed it.

## Tested:
Unit and E2E tests has passed.


adrianludwin:
Whoops, good catch!

/ok-to-test
/lgtm
/approve

Issue #276, opened 2023-05-07
Built-in propagation exceptions and the "propagate.hnc.x-k8s.io/all" annotation


duduedri96:
Version v1.1.0-rc2

I have a use case in which a Rancher cluster creates RoleBindings on my project-managed namespace, and I would like to have child namespaces to this parent namespace which will get those RoleBindings. The parent namespace is controlled by Rancher, and all role bindings should be propagated to the children.
The documentation states that Rancher resources that contain the `cattle.io/creator=norman` label are excluded by default. This prevents me from implementing my use case.

After diving a bit in the code, I figured out that if I'll have the `propagate.hnc.x-k8s.io/all=true` annotation on my resource it will be propagated, even with the default `Propagate` mode that RoleBindings get implicitly. I tested it out and it works. The reason it works is that the `ShouldPropagate` function checks for this `propagate.hnc.x-k8s.io/all=true` annotation before it figures out if the resource contains one of the excluded labels.

The documentation [clearly states](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#limit-the-propagation-of-an-object-to-descendant-namespaces) that the `propagate.hnc.x-k8s.io/all` annotation is useful only in the `AllowPropagate` propagation mode. This clearly isn't the case here, since RoleBindings use the `Propagate` mode.

The question is which part is wrong here - the documentation which should be updated to indicate this, or the code which shouldn't allow those resources to be propagated with the `propagate.hnc.x-k8s.io/all` annotation since they have a label which should be excluded by default? In other words, can I rely on this behavior to stay as is on future versions?

Thanks.

adrianludwin:
lol! That's a great corner case. On the one hand, the code is wrong - we never _intended_ to override the default exclusion. However, it looks like you've discovered a nice use case for the "bug" - it can override the default behaviour when that's not what you want.

What if we made this the official behaviour? That is, automatically-excluded objects are now only excluded _by default_, but you can override this with the annotation. Or would that not be useful enough?

I don't know much about Rancher. Can you explain why other Rancher users may have wanted this exclusion, and why this doesn't apply to your use case?

duduedri96:
Hi!
[...]

PR #280, opened 2023-05-21
Added CLI argument to disable default resource exclusion by label

duduedri96:
Currently Rancher-managed resources are skipped and excluded by default without a way to disable this behavior. This PR adds a CLI argument to disable the exclusion.

Ref. https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/104 https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/125 https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/276

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -55,28 +55,29 @@ var (
 )
 
 var (
-	probeAddr               string
-	metricsAddr             string
-	enableStackdriver       bool
-	maxReconciles           int
-	enableLeaderElection    bool
-	leaderElectionId        string
-	noWebhooks              bool
-	debugLogs               bool
-	testLog                 bool
-	internalCert            bool
-	qps                     int
-	webhookServerPort       int
-	restartOnSecretRefresh  bool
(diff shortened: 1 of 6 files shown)


adrianludwin:
Sorry it's taken me a while to review this.

Can you explain a bit more about why you might want these objects to be propagated? Do we need something more fine-grained - e.g. do we really want _all or none_ of the resources to be propagated? Thanks!

adrianludwin:
nvm, found #276 , will continue discussion there

duduedri96:
Commented on #276 with more details.

About your question, since the feature was added for excluding resources with Rancher annotations in mind, I thought it may be fit to exclude this feature altogether when the CLI argument is given (the mapping currently has only the Rancher annotation in it). Hence, I made the CLI flag to disable the resource exclusion by label altogether (i.e including future labels which may be added to this mapping). You think it makes more sense to have the CLI argument exclude only Rancher labels? To me it seems a bit too specific (feature-wise).

Thanks!

PR #283, opened 2023-05-26
Show HierarchicalResourceQuota status by kubectl-hns

mochizuki875:
## What type of PR is this?:
/kind feature

## What this PR does / why we need it:
As #274,  there is no way to know the quota and consumption of resource in the `HierarchicalResourceQuota` other than check the detail.
So I added new sub-command to show `HierarchicalResourceQuota` status as below.

```console
$ kubectl hns hrq -n test
NAME         AGE     REQUEST                                              LIMIT
test-hrq     2m37s   requests.cpu: 500m/1, requests.memory: 100Mi/200Mi   limits.cpu: 500m/2, limits.memory: 100Mi/500Mi
test-hrq-2   4m6s    requests.memory: 100Mi/300Mi                         limits.cpu: 500m/1

$ kubectl hns hrq test-hrq -n test
NAME       AGE     REQUEST                                              LIMIT
test-hrq   2m43s   requests.cpu: 500m/1, requests.memory: 100Mi/200Mi   limits.cpu: 500m/2, limits.memory: 100Mi/500Mi

$ kubectl hns hrq --all-namespaces
NAMESPACE   NAME         AGE     REQUEST                                              LIMIT
test        test-hrq     2m49s   requests.cpu: 500m/1, requests.memory: 100Mi/200Mi   limits.cpu: 500m/2, limits.memory: 100Mi/500Mi
test        test-hrq-2   4m18s   requests.memory: 100Mi/300Mi                         limits.cpu: 500m/1
```

## Tested:
Unit and E2E tests has passed.
I also added test case to HRQ E2E test which is now Pending.

## Which issue(s) this PR fixes:
#274

## Special notes for your reviewer:
In this case, if there is a `.status.hard.<requests/limits>.<resource>` field in `HierarchicalResourceQuota`, the corresponding content should appear in REQUEST and LIMIT.

I tried using `JsonPath` with `HierarchicalResourceQuota`'s `CustomResourceDefinition`(`.spec.versions.additionalPrinterColumns` field), but I couldn't come up with an expression to control what is displayed.

So in this PR, I propose a new sub-command of `kubectl-hns` in the following format to view the status of `HierarchicalResourceQuota`

```
kubectl-hns hrq [NAME] [flags]
```

If you have better ideas for doing that, please let me know and I'll fix it.

PR #286, opened 2023-05-28
Remove unneeded dependency to k8s.io/kubernetes

erikgb:
It seems like a dependency to `k8s.io/kubernetes` was added in https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/283. The Kubernetes project discourages importing `k8s.io/kubernetes`. Consumers should instead import the synced submodules. And if I am not overlooking anything, this dependency is actually not needed, as it is only used in dead code - which I have now removed.

The reason for looking into this is that I am currently working on a PR that will probably require a bump of controller-runtime. Bumping K8s dependencies are hard enough as it is. 😄 

/assign @adrianludwin


mochizuki875:
@erikgb!
As you say, `k8s.io/kubernetes` is not needed #283.
Thank you your fix!

adrianludwin:
Thanks for catching this @erikgb !

/lgtm
/approve

PR #287, opened 2023-05-28
Bump Kubernetes dependencies to 1.26.x

erikgb:
This PR bumps the K8s dependencies to 1.26.x. My main motivation is to get access to new features available in recent versions of controller-runtime. I have not bumped to the latest version of c-r, since it contains a lot of breaking changes that will require changing a lot more code, ref. https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0. I could eventually fix this in a follow-up PR.

As a consequence of the bump of K8s dependencies, it is required to bump other dependencies like cert-controller as well.

All tests run fine: normal tests and e2e-tests (also in `HNC_REPAIR` mode). I notice a few TLS errors in controller logs, but I do not know if this is caused by my corporate network environment. But I wouldn't mind someone else checking it!


adrianludwin:
Hey @erikgb is there some reason to bump this now?

erikgb:
@adrianludwin I tried to write tests in https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/285, and is blocked by features added to newer controller-runtime versions. I can always work around it, by copying code. Is there a reason for _not_ upgrading? 😉

adrianludwin:
Nope, that sounds great! Just wanted to know :) Thanks!

erikgb:
@adrianludwin Any idea what has happened to the CI? It ran yesterday....

config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -112,8 +111,8 @@ spec:
                   description: "Condition contains details for one aspect of the current
                     state of this API Resource. --- This struct is intended for direct
                     use as an array at the field path .status.conditions.  For example,
-                    type FooStatus struct{ // Represents the observations of a foo's
-                    current state. // Known .status.conditions.type are: \"Available\",
+                    \n type FooStatus struct{ // Represents the observations of a
```
adrianludwin, line 114: Any idea what this `\n` is?

adrianludwin:
No idea, seeing it on the postsubmit as well:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/hnc-postsubmit-test/1663185368003907584

config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
erikgb: I think it originates from the upstream K8s `Condition` type doc. It is what controller-gen "spits" out now.

adrianludwin:
Maybe file a bug to the Prow folks? Other repos seem to have passing tests today, so maybe it's just us? A _quick_ Google search didn't turn up any obvious problems.

config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
adrianludwin: That newline just looks weird, it's not showing up anywhere else? Maybe check the comment in the source file to make sure there's no weird newline there? But obviously not a big deal if it's not obvious.

erikgb: This is how the source looks like:

````go
// Condition contains details for one aspect of the current state of this API Resource.
// ---
// This struct is intended for direct use as an array at the field path .status.conditions.  For example,
//
//	type FooStatus struct{
//	    // Represents the observations of a foo's current state.
//	    // Known .status.conditions.type are: "Available", "Progressing", and "Degraded"
//	    // +patchMergeKey=type
//	    // +patchStrategy=merge
//	    // +listType=map
//	    // +listMapKey=type
//	    Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
//
//	    // other fields
//	}
type Condition struct {
````

It seems like the empty line adds `/n` to the CRD field description.

adrianludwin:
Great, glad to know it's not just us :)

PR #293, opened 2023-05-31
Changed default excluded labels from hardcoded values to be provided as CLI args, fixed incorrect exclusion bypass with annotation

duduedri96:
As discussed in #276, currently resources which are created by Rancher (and have the `cattle.io/creator=norman` label) are being excluded by default, without any way to disable this behavior. In addition, there a bug that causes resources with the `propagate.hnc.x-k8s.io/all` annotation to be propagated even though they are excluded.

This PR changes the exclude mechanism to rely on a CLI argument to the manager instead of a hardcoded value. It also sets the default arguments of the manager to have the `cattle.io/creator=norman` excluded by default, to keep the same default behavior as before. Documentation is updated to reflect the change. In addition, the bug that caused excluded resources to be propagated when using the `propagate.hnc.x-k8s.io/all` is fixed.

Edit: force-pushed an update that also fixes the bug that was mentioned in the issue.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,6 +18,7 @@ package main
 import (
 	"errors"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"strings"
@@ -72,6 +73,7 @@ var (
 	excludedNamespaces      arrayArg
 	managedNamespaceLabels  arrayArg
 	managedNamespaceAnnots  arrayArg
+	nopropagationLabel      arrayArg
 	includedNamespacesRegex string
 	webhooksOnly            bool
 	enableHRQ               bool
@@ -156,10 +158,31 @@ func parseFlags() {
 	flag.BoolVar(&enableHRQ, "enable-hrq", false, "Enables hierarchical resource quotas")
 	flag.StringVar(&hncNamespace, "namespace", "hnc-system", "Namespace where hnc-manager and hnc resources deployed")
 	flag.DurationVar(&hrqSyncInterval, "hrq-sync-interval", 1*time.Minute, "Frequency to double-check that all HRQ usages are up-to-date (shouldn't be needed)")
+	flag.Var(&nopropagationLabel, "nopropagation-label", "A label specified as key=val that, if present, will cause HNC to skip objects that match this label. May be specified multiple times, with each key=value pair specifying one label. See the user guide for more information.")
 	flag.Parse()
 
 	// Assign the array args to the configuration variables after the args are parsed.
 	config.UnpropagatedAnnotations = unpropagatedAnnotations
+
+	// Assign the exclusion label args to the configuration
+	for _, label := range nopropagationLabel {
+		keyVal := strings.Split(label, "=")
+		if len(keyVal) != 2 {
+			setupLog.Info(fmt.Sprintf("--nopropagation-label must contain exactly one equal sign (=): %s", label))
+			os.Exit(1)
+		}
+		labelKey := keyVal[0]
+		labelValue := keyVal[1]
+
+		if len(labelKey) == 0 {
+			setupLog.Info(fmt.Sprintf("--nopropagation-label must not have an empty key for the label: %s", label))
+			os.Exit(1)
+		}
+
+		setupLog.Info(fmt.Sprintf("Will exclude objects that match label %s", label))
+		config.NoPropagationLabels = append(config.NoPropagationLabels, config.NoPropagationLabel{Key: labelKey, Value: labelValue})
+	}
+
 	config.SetNamespaces(includedNamespacesRegex, excludedNamespaces...)
 	if err := config.SetManagedMeta(managedNamespaceLabels, managedNamespaceAnnots); err != nil {
 		setupLog.Error(err, "Illegal flag values")
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -49,6 +49,9 @@ spec:
         - "--excluded-namespace=kube-public"
         - "--excluded-namespace=hnc-system"
         - "--excluded-namespace=kube-node-lease"
+        # Preserves the existing behavior of skipping Rancher objects.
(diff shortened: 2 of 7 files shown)


duduedri96:
Note: force-pushed an update that also fixes the bug that was mentioned in the GitHub issue.

cmd/manager/main.go
```diff
@@ -156,10 +158,31 @@ func parseFlags() {
 	flag.BoolVar(&enableHRQ, "enable-hrq", false, "Enables hierarchical resource quotas")
 	flag.StringVar(&hncNamespace, "namespace", "hnc-system", "Namespace where hnc-manager and hnc resources deployed")
 	flag.DurationVar(&hrqSyncInterval, "hrq-sync-interval", 1*time.Minute, "Frequency to double-check that all HRQ usages are up-to-date (shouldn't be needed)")
+	flag.Var(&excludeResourcesWithLabel, "exclude-objects-with-label", "An annotation specified as key=val that, if present, will cause HNC to skip objects that match this label. May be specified multiple times, with each instance specifying one label. See the user guide for more information.")
```
adrianludwin, line 161: "An annotation" -> "A label"

"with each instance specifying one label" -> "one key/value pair" (or something like that)

Maybe change the name to `--nopropagation-label` to match the terminology [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#limit-the-propagation-of-an-object-to-descendant-namespaces)? And ditto for the internal names.

This will also make it more natural to document - e.g. "you can use these built-in annotations to disable propagation, _or_ if the objects already have the labels on them (e.g. because Rancher set them), the HNC admin can set this command-line arg."

config/manager/manager.yaml
```diff
@@ -49,6 +49,7 @@ spec:
         - "--excluded-namespace=kube-public"
         - "--excluded-namespace=hnc-system"
         - "--excluded-namespace=kube-node-lease"
+        - "--exclude-objects-with-label=cattle.io/creator=norman"
```
adrianludwin, line 52: Maybe add a comment above this line about why this is here - e.g. "preserve preexisting behaviour, please don't add any more to the default list of args"

docs/user-guide/how-to.md
```diff
@@ -956,3 +956,10 @@ Interesting parameters include:
   load on your metrics database (through increased metric cardinality) and also
   by increasing how carefully you need to guard your metrics against
   unauthorized viewers.
+* `--exclude-objects-with-label`: present by default with label
```
adrianludwin, line 959: I'd move the note about the Rancher stuff to a subbullet at the end. I'd start by saying it has the same effect as [disabling propagation](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#limit-the-propagation-of-an-object-to-descendant-namespaces) but is useful when you can't control the labels on the objects. And then just mention at the end that the default deployment of HNC excludes Rancher.

internal/config/default_config.go
```diff
@@ -8,3 +8,14 @@ package config
 // This value is controlled by the --unpropagated-annotation command line, which may be set multiple
 // times.
 var UnpropagatedAnnotations []string
+
+// ExclusionByLabelsSpec specifies a label Key and Value which will cause an object to be excluded
+// from propagation if the object defines that label with this specific value.
+type ExclusionByLabelsSpec struct {
```
adrianludwin, line 14: Suggestion: `NoPropagationLabel` and then the variable is simply `NopropagationLabels` or something like that.

internal/objects/reconciler_test.go
```diff
@@ -467,6 +472,18 @@ var _ = Describe("Basic propagation", func() {
 		Eventually(HasObject(ctx, "configmaps", barName, "cm-with-label-2")).Should(BeTrue())
 	})
 
+	It("should not propagate objects excluded by labels when using all annotation", func() {
```
adrianludwin, line 475: This is a change from the current behaviour - why not allow the exclusion to be overridden?

Either way, can you please document this as well?

adrianludwin (commented):
Great change! Just a few comments, mainly aesthetic but one functional.

duduedri96:
Thanks @adrianludwin ! Will work on that during the next weekend.

internal/objects/reconciler_test.go
duduedri96: Hmm the thing is that it's indeed a change in behavior, however as we discussed on the original issue, it was never intended. In addition, the current documentation never mentioned that it can work this way - it only says that the 'all' annotation is useful on the AllowPropagate mode. I actually discovered this behavior by looking at the code. What are your thoughts here? Any suggestion as to how we should document it further than what I mentioned?

duduedri96:
@adrianludwin , I made the changes you requested, can you please re-review?

About the behavior change, I left it as-is and replied on your comment above with my thoughts, and I'm not sure what to do further.
Thanks!

internal/objects/reconciler_test.go
rjbez17: I tend to agree with @duduedri96 but do understand this is a breaking change so we may want to consider not breaking the behavior. I guess I'm saying I could go either way here :).

adrianludwin: Ok, I'm good to leave the (new) behaviour as-is, and yes, I don't think we need any more documentation unless someone asks for it.

adrianludwin (commented):
lgtm after a few final changes, thanks!

duduedri96:
@adrianludwin , implemented your suggestions.
Once you set lgtm, should I assign tashimi for approval (as the ci-robot suggested) or you'll approve?

Thank you!

adrianludwin:
@duduedri96 thanks for this change! Can you please cherry-pick it into the v1.1 branch as well? Then I'll release v1.1 RC2 for testing.

duduedri96:
Thanks @adrianludwin , created #299  cherry pick PR.

PR #290, opened 2023-05-30
Cherry pick of #283, #286, #294

keisukesakasai:
Cherry pick of https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/283 , https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/286 , https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/294 on master.

https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/283: Show HierarchicalResourceQuota status by kubectl-hns
https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/286 : Remove unneeded dependency to k8s.io/kubernetes
https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/294: Remove unneeded replace clause from go.mod

The discussion regarding the content can be found here:
https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/284#issuecomment-1566706843

PR #294, opened 2023-06-05
Remove unneeded replace clause from go.mod

mochizuki875:
Remove unneeded replace clause from `go.mod`.
It was added as workaround of `k8s.io/kubernetes` import error in #283 .
(ref. https://github.com/kubernetes/kubernetes/issues/79384)

However, `k8s.io/kubernetes` has been removed in #286 as unneeded dependency so replace clause is unneeded too.

related: https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/290#pullrequestreview-1457869921


erikgb:
Thanks!
/lgtm

PR #295, opened 2023-06-07
Show HierarchicalResourceQuota status by kubectl get

mochizuki875:
## What type of PR is this?:
/kind feature

## What this PR does / why we need it:
We have added `kubectl-hns hrq` sub-command to show `HierarchicalResourceQuota` status in easy way.(#283)
Hoewver, it's not yet possible to achive that using the Kubernetes standard way `kubectl get hrq`.

```console
$ kubectl get hierarchicalresourcequotas -n test
NAME       AGE
test-hrq   80s
```

So in this PR, I made it possible to show `HierarchicalResourceQuota` status by using `kubectl get hrq`.

```console
$ kubectl get hrq -n test
NAME         REQUEST                                              LIMIT
test-hrq     requests.cpu: 500m/1, requests.memory: 100Mi/200Mi   limits.cpu: 500m/2, limits.memory: 100Mi/500Mi
test-hrq-2   requests.memory: 100Mi/300Mi                         limits.cpu: 500m/1

$ kubectl get hrq --all-namespaces
NAMESPACE   NAME         REQUEST                                              LIMIT
test-2      test-hrq-3   requests.cpu: 0/4, requests.memory: 0/500Mi          limits.cpu: 0/8, limits.memory: 0/750Mi
test        test-hrq     requests.cpu: 500m/1, requests.memory: 100Mi/200Mi   limits.cpu: 500m/2, limits.memory: 100Mi/500Mi
test        test-hrq-2   requests.memory: 100Mi/300Mi                         limits.cpu: 500m/1
```

## Tested:
Unit and E2E tests has passed.
I also added test case to HRQ E2E test which is now Pending.

## Related:
#274
#283
#289

## Special notes for your reviewer:
`HierarchicalResourceQuota` documentation is proceeding by @zfrhv and @adrianludwin.(#289) 
If this PR will be merged to `v1.1`, we should add this feature to it.

PR #300, opened 2023-06-16
Fix generate HRQ CRD from markers

erikgb:
While working on https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/285, I noticed that running `make manifests` updates the HRQ CRD. This also happens on the master branch, which I think is not intentional and wrong. This PR fixes this by regenerating the HRQ CRD.

The PR also marks the current HRQ as the storage version - which I discovered as missing when comparing the kubebuilder markers for the CRDs.

I think the difference is caused by a minor bug in `controller-gen`, and is related to the recently added print columns on HRQ. The columns include fields from the HRQ **status**, and I think there is some code in controller-gen assuming the CRD has a status subresource when including fields from status as print columns.

I really think we should have status as subresources - especially on HRQ, but I don't have the bandwidth to fix that.

/cc @adrianludwin

### DIFF
--- a/api/v1alpha2/hierarchicalresourcequota_types.go
+++ b/api/v1alpha2/hierarchicalresourcequota_types.go
@@ -50,6 +50,7 @@ type HierarchicalResourceQuotaStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:path=hierarchicalresourcequotas,shortName=hrq,scope=Namespaced
+// +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="Request",type="string",JSONPath=".status.requestsSummary"
 // +kubebuilder:printcolumn:name="Limit",type="string",JSONPath=".status.limitsSummary"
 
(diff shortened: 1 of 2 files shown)


mochizuki875:
@erikgb 

> I think the difference is caused by a minor bug in controller-gen, and is related to the recently added print columns on HRQ. 

Thank you for your fix. It may have been caused by my fix #295 ...

> I really think we should have status as subresources - especially on HRQ, but I don't have the bandwidth to fix that.

As you may know that, if we make status subresource, we can independently update status field using [StatusClient](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client#StatusClient) which is member of [client.Client](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/client#Client).
(ex. we can update only status field using `r.Status().Update()`.)

Although this use case dosn't seems to exist in HNC, [it is generally recommended](https://book.kubebuilder.io/reference/generating-crd.html#subresources) to make status subresource.
So it might be good to do it.

erikgb:
@mochizuki875 We use the status subresource in all other controllers I am involved in, so HNC is quite special in this sense. I have earlier looked at doing this in HNC, and I think it is not a simple task..... But good luck if you make an attempt! 👍

mochizuki875:
@erikgb 
That's for sure.
Although I don't think it's so important task at this point, I'll check the implementation and if it looks doable, I'll do it:)

rjbez17:
/lgtm

@adrianludwin ?

PR #302, opened 2023-06-21
Fix Quickstart Document about allowCascadingDeletion

mochizuki875:
## What type of PR is this?:
/kind bug
/kind documentation

In _Subnamespaces deep dive_ section of Quickstart Document, it will fail to delete `subnamespaceanchors/service-4` because it will not be set cascading deletion explicitly set(`--allowCascadingDeletion`).
So I've fixed this in this PR.

https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/564b4a968ca2091f27cf0eb51d0cd99190e57db5/docs/user-guide/quickstart.md?plain=1#L678-L690

## Tested:
Unit and E2E tests has passed.
In E2E tests, `--allowCascadingDeletion` has already been set to `subnamespaceanchors/service-4` so it will sucess.

https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/564b4a968ca2091f27cf0eb51d0cd99190e57db5/test/e2e/quickstart_test.go#L280-L287

### DIFF
--- a/docs/user-guide/quickstart.md
+++ b/docs/user-guide/quickstart.md
@@ -623,7 +623,7 @@ subnamespaces by default.
 
 HNC will only allow recursive deletion of subnamespaces if those subnamespaces,
 or any of their ancestors, have cascading deletion explicitly set. For example,
-to delete `service-1`, we first enable cascading deletion and then remove it:
+to delete `service-1`, you first enable cascading deletion and then remove it:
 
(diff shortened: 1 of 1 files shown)


mochizuki875:
@adrianludwin 
Hi, thanks for your feedback!
I've fixed it and related explanation about subnamespace.
How about that?

PR #311, opened 2023-07-02
kubectl-hns: Add "delete" command

### LINKED ISSUE #307: `kubectl hns delete` command?
While `kubectl hns create` command exists, `hns delete` command does not.

We can delete subns with `kubectl delete subns` command. However, a user using `kubectl hns` command for the first time will assume that since `create` exists, `delete` must exist, but will be confused because it does not.

superbrothers:
This PR adds "delete" command that deletes subnamespaces to the kubectl-hns plugin. The command has `--allowCascadingDeletion` flag to allow cascading deletion of its subsnamespaces.

Closes #307

### DIFF
--- a/docs/user-guide/how-to.md
+++ b/docs/user-guide/how-to.md
@@ -236,6 +236,12 @@ Instead, you must delete its anchor (note that `subns` is a short form of
 $ kubectl delete subns child -n parent
 ```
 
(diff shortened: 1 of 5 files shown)


rjbez17:
/ok-to-test
/lgtm

@adrianludwin this lgtm, LMK if you want to take a look.

adrianludwin:
/lgtm
/approve

Thanks for this!

PR #298, opened 2023-06-11
Documentation for #295

zfrhv:
Added documetation for:
#295: `kubectl get hrq` command

And also corrected a little the files.

@mochizuki875 do you want me to add something else?
Turned out to be shorter than I expected.

PR #289, opened 2023-05-29
HierarchicalResourceQuota documentation

zfrhv:
solves #264 

i have a feeling my charges are not that great,
@adrianludwin what to you think of it for now?

PR #315, opened 2023-07-04
docs: improve grammar and simplify HRQ examples 

agilgur5:
## Summary

Was reading through the docs and these seemed the most verbose
- it seems to be a very recent feature as well, from #289

## Details

- use same heading in both Quickstart and Concepts, including HRQ abbreviation
  - use abbreviation to shorten the HRQ docs themselves as well

### Grammar

- fix run-on sentences by adding punctuation
  - "by default, to use it" -> "by default. To use it"
  - and a few others
- add missing commas
  - "Then,", "finally,", etc
- add Oxford commas
  - ", but", ", and", ", or", etc

### Simplifications

- simplify several sentences, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - "won't be over the amount of resources that is configured in `HierarchicalResourceQuota`" -> "cannot surpass the HRQ"
    - "can't go over what is configured in the HRQ" -> "can't surpass the HRQ"
  - "And put the resources HRQ in" -> "An HRQ in"
    - "This will restrict whole" -> "restricts" (also uses present tense now, per style guide)
  - "the amount of resources that" -> "the resources"
    - "resource usage" -> "resources"
  - "automatically creates" -> "creates"

- in a similar vein, remove or replace ambiguous language
  - "equally shared" -> "shared" -- they can be shared however, not necessarily "equally"
  - "which is very efficient" -> "" -- "efficient" could be interpreted in a performance sense and is unnecesary
  - "lower level", "level" -> "child" / "sub" -- use HNC's existing terminology instead of "lower level", which has multiple disambiguations
    - Similarly use HNC's existing terminology for:
      - "individual", "members" -> "namespace", "subnamespaces"
      - "above", "overall" -> "parent"
  - "wins" -> "applies" -- replace idiom, per [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-jargon-and-idioms)
  - also remove the word "simply", per [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding)

- clarify a few sentences with a single word
  - "within", "Instead", "still", "directly"

- consolidate the two paragraphs on command-line usage into one
  - looks like they were added at different times, in #298
  - they largely duplicated each other with slightly different wording
  - creating an HRQ via a CR YAML is the definition of how CRs work, which are already used across HNC, so not necessary to re-explain

## Review Notes

There are potentially more changes that could be made according to the style guide, but I primarily focused changes on simplifications, clarifications, and grammar
- **EDIT**: See #316 as one follow-up

## Misc

This is a great example! It gives very practical, real-world use-cases which is very helpful for context and understanding how one might use HNC!

### DIFF
--- a/docs/user-guide/concepts.md
+++ b/docs/user-guide/concepts.md
@@ -227,18 +227,18 @@ You can create a subnamespace from the command line via `kubectl hns create chil
 
 ***Hierarchical resource quotas are beta in HNC v1.1***
 
-***HRQs are not included by default, to use it install the hrq.yaml file in the [releases -> Assets](https://github.com/kubernetes-sigs/hierarchical-namespaces/releases)***
+***HRQs are not included by default. To use it, install the `hrq.yaml` file in [Releases -> Assets](https://github.com/kubernetes-sigs/hierarchical-namespaces/releases)***
 
 When you want to give some amount of resources to `team-a`, and want them to be able to
 flexibly use resources in any of their subnamespaces, you create a `HierarchicalResourceQuota`
-in namespace `team-a`. The sum of all resources from all the subnamespaces of the
-members won't be over the amount of resources that is configured in
-`HierarchicalResourceQuota` of namespace `team-a`. All of the resources of `team-a` are
-equally shared between the applications in their subnamespaces, which is very efficient.
+in namespace `team-a`. The sum of all resources from all subnamespaces of `team-a`
+cannot surpass the HRQ in namespace `team-a`. All of the resources of `team-a` are
+shared between applications in their subnamespaces.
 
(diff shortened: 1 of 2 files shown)


adrianludwin (commented):
Thanks for this change!

/lgtm
/approve

adrianludwin:
Hi @agilgur5  , sorry for the long delay, but could you please rebase this change? I think the tests will pass after that's done. Thanks!

agilgur5:
@adrianludwin rebased. It needs to be re-approved now

Issue #317, opened 2023-07-17
Default subnamespace permissions 


zfrhv:
Each time when I install HNC I give permissions to the `clusterRole/edit` on the hnc custom resources.
It allows the clients create their own sub namespaces if they have at least edit role:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
  name: hnc-edit
rules:
- apiGroups:
  - hnc.x-k8s.io
  resources:
  - hierarchyconfigurations
  - hierarchicalresourcequotas
  - subnamespaceanchors
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - hnc.x-k8s.io
  resources:
  - hierarchicalresourcequotas
  - subnamespaceanchors
  verbs:
  - create
  - update
  - patch
  - delete
```

So I'm wondering, should I create a PR so it would be like that by default?

erikgb:
Yes, I think it makes sense. Maybe add permissions to the other default aggregated cluster roles (view, admin and cluster-reader) also?

zfrhv:
merged

PR #318, opened 2023-07-19
Add permissions on hnc CRDs to default clusterRoles: cluster-reader, view, edit, admin

zfrhv:
Changing permissions to default clusterRoles.
Each user that has admin/edit can manage sub-namespaces and HRQs
Each user that has view/cluster-reader can see all hnc CRs

Related to: #317


zfrhv:
Also does hnc supposed to have cluster-admin?:
https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/config/rbac/role.yaml#L19C1-L24C8

config/rbac/aggregate_to_view.yaml
```diff
@@ -0,0 +1,16 @@
+# Automatically aggregated to clusterRoles: cluster-reader, view -> edit -> admin
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
```
jouve, line 6:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-view: "true"
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

config/rbac/aggregate_to_view.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: hnc-view
```
adrianludwin, line 7: I don't *think* you need the `hnc-` prefix here, shouldn't Kustomize add that for you?

config/rbac/aggregate_to_edit.yaml
```diff
@@ -0,0 +1,18 @@
...
+- apiGroups:
+  - hnc.x-k8s.io
+  resources:
+  - hierarchicalresourcequotas
+  - subnamespaceanchors
+  verbs:
+  - create
+  - update
+  - patch
+  - delete
```
adrianludwin, line 18: Should this be '*'? I'm not 100% confident this is the full list :)

adrianludwin (commented):
If I'm understanding this correctly, nothing should change for "admin" users, but cluster-edit users will get the same permissions that "admins" used to have, and cluster-viewers will get "view" permissions. Is that correct? If so, this lgtm after some nits

adrianludwin:
> Also does hnc supposed to have cluster-admin?: https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/config/rbac/role.yaml#L19C1-L24C8

Basically, yes, since HNC can be used to copy any namespaced object. If there were a way to only give permissions for namespaced objects we'd do that, but AFAIK this doesn't exist.

Individual HNC users can easily restrict HNC to only have access to an allowlisted set of resources, but this is the only default that will reliably work in all cases.

zfrhv:
> If there were a way to only give permissions for namespaced objects we'd do that, but AFAIK this doesn't exist.

Isn't thats what the operatorGroups are made for?  
When installing an operator with OLM operatorGroup must exist, and there you can specify the list of the namespaces on which the operator will have permissions.  
You can even use a [label selector](https://docs.openshift.com/container-platform/4.8/operators/understanding/olm/olm-understanding-operatorgroups.html), just put the label on the namespace and the operator should get control of it.
I hope I'm not mistaken.

adrianludwin:
OperatorGroups appear to only be an OpenShift concept, not a K8s concept. So I can't comment on that... plus from your description it's a way to select the _namespaces_ for the operator, but not the _types of objects_.

zfrhv:
> OperatorGroups appear to only be an OpenShift concept

OLM can be installed in kubernetes, but I understand if you dont want to use it because probably most of kubernetes people dont use it.

> plus from your description it's a way to select the namespaces for the operator, but not the types of objects.

Im pretty sure you can configure the objects that you need the permissions for, and OLM will create permissions for those objects in the listed namespaces (maybe you can specify `*`)

I'm not too confident because the only operator I made was from helm chart, and it was bad quality.

config/rbac/aggregate_to_edit.yaml
jouve: you can find the verbs with `kubectl api-resources -o wide`:
```console
% kubectl api-resources -o wide --namespaced --api-group=hnc.x-k8s.io
NAME                         SHORTNAMES   APIVERSION              NAMESPACED   KIND                        VERBS                                                        CATEGORIES
hierarchicalresourcequotas   hrq          hnc.x-k8s.io/v1alpha2   true         HierarchicalResourceQuota   delete,deletecollection,get,list,patch,create,update,watch   
hierarchyconfigurations                   hnc.x-k8s.io/v1alpha2   true         HierarchyConfiguration      delete,deletecollection,get,list,patch,create,update,watch   
subnamespaceanchors          subns        hnc.x-k8s.io/v1alpha2   true         SubnamespaceAnchor          delete,deletecollection,get,list,patch,create,update,watch   
```

so
```suggestion
  - create
  - update
  - patch
  - delete
  - deletecollection
```

config/rbac/aggregate_to_view.yaml
zfrhv: Maybe yes, I didnt saw prefix configuration in local kustomization file, so idk.  
I will remove it tho

config/rbac/aggregate_to_edit.yaml
zfrhv: yes, I wasnt sure if in edit I should put only the edit permissions and relay on the aggregation from view, or should I put *.  
I will change it to '*' for the simplicity

config/rbac/aggregate_to_edit.yaml
```diff
@@ -0,0 +1,15 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: edit
+rules:
+- apiGroups:
+  - hnc.x-k8s.io
+  resources:
```
jouve, line 11: current admin-role has `*` so `hncconfigurations` too

config/rbac/aggregate_to_edit.yaml
jouve: k8s [system:aggregate-to-edit](https://github.com/kubernetes/kubernetes/blob/0241da314e0e69817d66313b45a69c19d1ce7327/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L651) uses explicit verbs:  with [Write](https://github.com/kubernetes/kubernetes/blob/0241da314e0e69817d66313b45a69c19d1ce7327/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L35)

zfrhv: Yes so I wasn't sure if to put Write + relay on view aggregation, or ReadWrite.  
I think ReadWrite is better because then this ClusterRole can be used for other purposes and not only aggregation to edit.  
And ReadWrite is `'*'` anyways

config/rbac/aggregate_to_edit.yaml
zfrhv: why hncconfigurations too?
isnt its clusterWide object?
you mean `hierarchyconfigurations`?

jouve: indded :)

zfrhv: I dont have practice with `hierarchyconfigurations` so I dont know the consequences, a bit afraid to add something I dont know

zfrhv: Ok from what I see this CR allows to add annotations/labels to current namespace and his children.  
This roleBinding is aggregated to edit/admin permission, user that has edit/admin shouldn't be able to modify the namespace including annotations/labels.  

If `hierarchyconfigurations` would influence only the children namespaces then yes I would add it, but because it effects the parent namespace as well, I consider it as a security breach because edit/admin on the namespace shouldnt have permissions to modify the parent namespace.

Sorry I wont add it to the list, if there will be a way to configure `hierarchyconfigurations` only to effect the children (maybe from `hncConfiguration` or something) then yes we can add it as default edit permission.

But thank you for the feedback!

jouve: then you need to add back the admin-role which [included it and aggregated it to admin](it https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/318/files#diff-bc466b58edebab97ebe341a6e371ee1c6ac9a93ce2b783c90c3e267c05284f8cL6-L18)

zfrhv: I see, so it was already configured like this before, thank you for mentioning

@adrianludwin why user with admin role has permissions to edit the namespace labels/annonations?
By default admin cant change anything about the namespace object.

zfrhv:
Also I didn't knew i had to update incubator/hnc/config/rbac/kustomization.yaml
so imma do that

config/rbac/aggregate_to_edit.yaml
adrianludwin: It might make sense to distinguish between `hnc-edit`, which gives permission to the namespaced objects (`SubnamespaceAnchor`, `HierarchyConfiguration`, `HierarchicalResourceQuota`) and `hnc-admin`, which would add permission to administer HNC itself (`HNCConfiguration`). That is, `hnc-edit` would let users modify any part of the namespace hierarchy, while `hnc-admin` would let users change the way HNC works.

Being able to edit `HierarchyConfiguration` _looks_ like it lets you change the parent of a namespace, but in reality [HNC itself imposes limitations on this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/concepts.md#namespaces-administrators):

> However, being the admin of a namespace does not give you free reign over that namespace’s config. For example, if you are the admin of a child namespace but not its parent, you may not change the parent of your namespace, as this would remove privileges from the admin of the parent (otherwise known as a superadmin, which in HNC's context means an admin of any ancestor namespace). Similarly, even if your namespace is a root, you may not set its parent to any namespace of which you are not currently an admin, since the namespace could then inherit sensitive information such as Secrets.
> 
> In general, to change the parent of your namespace N from A to B, you must have the following privileges:
> 
> * You must be the admin of the highest namespace that will no longer be an ancestor of namespace N after this change, in order to confirm that you are happy to lose your privileges in namespace N.
> * You must be the admin of namespace B, in order to acknowledge that sensitive objects from B may be copied into namespace N.

This is implemented [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/74fe01460083fd410a8fbd728eb255c4d728dbcb/internal/hierarchyconfig/validator.go#L304).

jouve: admin is supposed to be used with a RoleBinding [ref](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles), so a user would not be able to use HNCConfiguration which is ClusterScoped anyway:
* view: view resources in namespace
* edit: view+edit resources (except rbac Role/Rolebinding) in namespace
* admin: view+edit resource (incl. Role/RoleBinding) in namespace
* cluster-admin: view+edit anything (resources `*` verbs: `*`) [ref](https://github.com/kubernetes/kubernetes/blob/82bca6304b5365f7df2627ad2a6fb3d4539bf36f/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L199C19-L200) , with clusterrolebinding

zfrhv: Sounds great, but my point was that without HNC, when giving someone admin on namespace A, he cannot change the labels nor annotations of that namespace.
But with the help of HNC, the admin user can configure hiearchyConfiguration to add annotations including on self namespace A.

So its kind of a workaround to be able to manage the namesoace labels+annotations without actually having the permissions.

But now as I think, why not to actually give the ability for the user to label or annotate his namespace even if he cant modify the object? (as long as he is admin)
I bet its not allowed as default only because labels+annotations are part of the object, and its impossible to give permissions only on a specific fields in the object.

zfrhv: > But yes, granting someone hnc-admin at the cluster level would allow them to modify any managed label/annotation on any namespace; that's by design.

Thats the point, if someone is granted at not cluster level (only namespaced) he can still modify the annotations of his namespace.

Simple user has only roleBinding and can create hierarchyConfiguration, HNC will create the annotation for him even if he doesn't has permissions.

Anyways I think if someone has admin on the namespace it makes sense that he could add labels to the namespace.

Give me 30 min I will update the PR

adrianludwin:
Failure looks unrelated, probably just a flake?

/lgtm
/retest

PR #321, opened 2023-07-21
Support unpropagated labels

timbyr:
By default ArgoCD use labels to understand which objects it controls.

When HNC propagates those resources ArgoCD assumes ownership of those
resources and attempts to reconcile it with current state, usually
leading to deletion of those resources.
This causes a feedback loop where ArgoCD requests a deletion of an
object, that is blocked by the HNC webhook, and HNC no longer is able to
properly manage the resource due to the deletion finaliser set on the
resource.

This is similar behaviour to ConfigSync with annotations.

This commit adds an unpropagated-labels command line argument following the
implementation of the unpropagated-annotations argument. This allows
cluster operators to disable propagation of labels owned by systems like
ArgoCD.

Tested:
* Implemented additional e2e test matching unpropagated-annotations tests.
* Manually validated that configured labels were not propogated as expected
  and ArgoCD did not adopt propgated resources.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -70,6 +70,7 @@ var (
 	webhookServerPort       int
 	restartOnSecretRefresh  bool
 	unpropagatedAnnotations arrayArg
+	unpropagatedLabels      arrayArg
 	excludedNamespaces      arrayArg
 	managedNamespaceLabels  arrayArg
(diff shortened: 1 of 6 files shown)


adrianludwin (commented):
Sorry to leave this so long, but this is a great change, no notes! Thanks for the submission :)

/lgtm
/approve

Issue #347, opened 2023-11-15
Use the default namespace in the kubeconfig


superbrothers:
hns command will fail to execute if the `-namespace` option is not specified.

```
$ kubectl hns create foo
Error: parent must be set via --namespace or -n
```

On the other hand, kubectl uses the default namespace set in kubeconfig if not specified. I think hns command's behavior should be aligned with kubectl's.

What do you think?

yanggangtony:
+1
Make sense for me.

superbrothers:
I'll make a PR.

yanggangtony:
I have done it. wait.

superbrothers:
OK. Thanks.

PR #348, opened 2023-11-15
Add default for the --namespace when it is not specify.

yanggangtony:
Fixes [issue 347](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/347).
Add default for the --namespace when it is not specify.

Before is:
image

After update :
image

### DIFF
--- a/internal/kubectl/root.go
+++ b/internal/kubectl/root.go
@@ -62,6 +62,10 @@ func init() {
 	client = &realClient{}
 
 	kubecfgFlags := genericclioptions.NewConfigFlags(false)
+	defaultNs, _, err := kubecfgFlags.ToRawKubeConfigLoader().Namespace()
+	if err != nil {
+		fmt.Printf("Error reading namespace in default kubeconfig :%s\n", err.Error())
+	}
 
 	rootCmd = &cobra.Command{
 		// We should use "kubectl hns" instead of "kubectl-hns" to invoke the plugin.
@@ -102,8 +106,8 @@ func init() {
 	rootCmd.AddCommand(newSetCmd())
 	rootCmd.AddCommand(newDescribeCmd())
 	rootCmd.AddCommand(newTreeCmd())
-	rootCmd.AddCommand(newCreateCmd())
-	rootCmd.AddCommand(newDeleteCmd())
+	rootCmd.AddCommand(newCreateCmd(defaultNs))
+	rootCmd.AddCommand(newDeleteCmd(defaultNs))
 	rootCmd.AddCommand(newConfigCmd())
 	rootCmd.AddCommand(newVersionCmd())
 	rootCmd.AddCommand(newHrqCmd())
--- a/internal/kubectl/create.go
+++ b/internal/kubectl/create.go
@@ -37,7 +37,7 @@ var createCmd = &cobra.Command{
 	},
 }
 
-func newCreateCmd() *cobra.Command {
-	createCmd.Flags().StringP("namespace", "n", "", "The parent namespace for the new subnamespace")
+func newCreateCmd(defaultNs string) *cobra.Command {
+	createCmd.Flags().StringVarP(&namespace, "namespace", "n", defaultNs, "The parent namespace for the new subnamespace")
 	return createCmd
 }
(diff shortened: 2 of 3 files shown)


superbrothers:
@yanggangtony This is not what is expected, as it uses the "default" namespace if `--namespace` is not specified. #347 says it wants to use the namespace specified in kubeconfig. This change also differs from the behavior of kubectl.

yanggangtony:
> @yanggangtony This is not what is expected, as it uses the "default" namespace if `--namespace` is not specified. #347 says it wants to use the namespace specified in kubeconfig. This change also differs from the behavior of kubectl.

you means 
```
contexts:
- context:
    cluster: development
    namespace: frontend
    user: developer
  name: dev-frontend
```
The ```namespace : frontend ``` for the current context used?

superbrothers:
> The `namespace : frontend ` for the current context used?

Yes.

yanggangtony:
> > The `namespace : frontend ` for the current context used?
> 
> Yes.

@superbrothers 
Thanks for quick review.
I will update later.

yanggangtony:
@superbrothers 
Like the test pic above , is that ok for this update?

superbrothers:
Yes, looks good. But the namespace specified for the current context will be changed. It is not a good idea to set a value that can be changed to a default value (I do not know the reviewer's opinion).

```diff
diff --git a/internal/kubectl/create.go b/internal/kubectl/create.go
index 314f2dd8..41500eba 100644
--- a/internal/kubectl/create.go
+++ b/internal/kubectl/create.go
@@ -16,9 +16,6 @@ limitations under the License.
 package kubectl

 import (
-       "fmt"
-       "os"
-
        "github.com/spf13/cobra"
 )

@@ -29,8 +26,7 @@ var createCmd = &cobra.Command{
        Run: func(cmd *cobra.Command, args []string) {
                parent, _ := cmd.Flags().GetString("namespace")
                if parent == "" {
-                       fmt.Println("Error: parent must be set via --namespace or -n")
-                       os.Exit(1)
+                       parent = defaultNamespace
                }
                // Create the anchor, the custom resource representing the subnamespace.
                client.createAnchor(parent, args[0])
diff --git a/internal/kubectl/root.go b/internal/kubectl/root.go
index 8e35e1ec..4a0a451b 100644
--- a/internal/kubectl/root.go
+++ b/internal/kubectl/root.go
@@ -41,6 +41,7 @@ var k8sClient *kubernetes.Clientset
 var hncClient *rest.RESTClient
 var rootCmd *cobra.Command
 var client Client
+var defaultNamespace string

 type realClient struct{}
 type anchorStatus map[string]string
@@ -77,6 +78,11 @@ func init() {
                                return err
                        }

+                       // Get the namespace set to current context in kubeconfig
+                       if defaultNamespace, _, err = kubecfgFlags.ToRawKubeConfigLoader().Namespace(); err != nil {
+                               return err
+                       }
+
                        // create the K8s clientset
                        k8sClient, err = kubernetes.NewForConfig(config)
                        if err != nil {
```

internal/kubectl/root.go
```diff
@@ -62,6 +62,10 @@ func init() {
 	client = &realClient{}
 
 	kubecfgFlags := genericclioptions.NewConfigFlags(false)
+	defaultNs, _, ise := kubecfgFlags.ToRawKubeConfigLoader().Namespace()
```
rjbez17, line 65: Can you update `ise` to `err` to follow the pattern of the rest of the repo.

rjbez17 (commented):
I'm on the fence with this one. It is nice to allow tools like `KNS` to work in general, but part of the requirement of the flag was to prevent accidental creation.

@adrianludwin any thoughts?

internal/kubectl/root.go
yanggangtony: yes, rebased.

yanggangtony:
Hey, friendly ping .
This pr is for a long time .
Would you put it for progress?

@rjbez17 @adrianludwin @srampal

yanggangtony:
Thanks.

yanggangtony:
thanks.

Issue #319, opened 2023-07-21
hns CLI: describe command should also describe allowCascadingDeletion property


manjinder-mckc:
**Current Output:**
```
$ kubectl hns describe boss                             
Hierarchy configuration for namespace boss
  Parent:  bigboss
  Children:
  - workers (subnamespace)
  No conditions
```
**Expected Output:**
```
$ kubectl hns describe boss                             
Hierarchy configuration for namespace boss
  Parent:  bigboss
  Children:
  - workers (subnamespace)
  AllowCascadingDeletion: true
  No conditions
```

Proposing Changes here: https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/564b4a968ca2091f27cf0eb51d0cd99190e57db5/internal/kubectl/describe.go#L33-L83

Would be willing to raise a PR, if you guys agree that this makes sense.

adrianludwin:
I'm not seeing the code that actually prints out `AllowCascadingDeletion` but that feature does look like a good idea, I'll review it if you send me a PR, thanks! One minor nit, maybe use human words, e.g. `Allows cascading deletion` instead of `AllowCascadingDeletion`.

Issue #303, opened 2023-06-22
Enable to know allowCascadingDeletion status by user friendly way


mochizuki875:
## What happened?
Trying [subnamespaces-deep-dive section of Quickstart](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md#subnamespaces-deep-dive), I think it's not user friendly that we can know `allowCascadingDeletion` status only by checking `HierarchyConfiguration`.(Although `ActivitiesHalted (ParentMissing)` can be checked by `kubectl hns describe`)

## What did you expect to happen?
I think we need other means to know `allowCascadingDeletion` status.

ex1)

Show the result of `kubectl hns tree`.

```console
$ kubectl hns tree team-a
team-a
├── [s] service-1 (AllowCascadingDeletion)
│   └── [s] dev
├── [s] service-2
├── [s] service-3
└── [s] service-4
    └── staging

[s] indicates subnamespaces
```

ex2)

Show the result of `kubectl get subns`.

```bash
$ kubectl -n team-a get subns
NAME        AGE         AllowCascadingDeletion
service-1   12m         true
service-2   5h31m
service-3   12m
service-4   8m20s
```

ex3)

Show the result of `kubectl hns describe`.
It is the same way of `ActivitiesHalted (ParentMissing)`.

```bash
$ kubectl hns describe service-1
Hierarchy configuration for namespace service-1
  Parent: team-a
  Children:
  - dev (subnamespace)
  Conditions:
  - AllowCascadingDeletion: true

No recent HNC events for objects in this namespace
```

## How can we reproduce it (as minimally and precisely as possible)?

Now, I know no other means other than the following:
(It's risky status of subnamespace so should be easy for users to notice)

```bash
$ kubectl get hierarchyconfiguration hierarchy -o jsonpath='{.spec}' -n service-1
{"allowCascadingDeletion":true,"parent":"team-a"}
```

or

```bash
$ kubectl describe hierarchyconfiguration hierarchy -n service-1
Name:         hierarchy
Namespace:    service-1
Labels:       <none>
Annotations:  <none>
API Version:  hnc.x-k8s.io/v1alpha2
Kind:         HierarchyConfiguration
Metadata:
  Creation Timestamp:  2023-06-22T06:21:52Z
  Finalizers:
    hnc.x-k8s.io/hasSubnamespace
  Generation:        3
  Resource Version:  137989
[...]

PR #355, opened 2024-01-30
feat: kubectl-hns describe to show AllowCascadingDeletion flag

manjinder-mckc:
Fixes Issues: #319 #303 

Enhances describe command in kubectl hns plugin to also show AllowCascadingDeletion flag.

```
$ kubectl hns describe boss                             
Hierarchy configuration for namespace boss
  Parent:  bigboss
  Children:
  - workers (subnamespace)
  Allows Cascading Deletion: true
  No conditions
```
*Tested:* Added test; verified that the test failed before the change and passed after it; ran e2e tests; also tested by hand on local cluster.

### DIFF
--- a/internal/kubectl/describe.go
+++ b/internal/kubectl/describe.go
@@ -25,7 +25,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	"k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


rjbez17:
Thanks for this, great addition

PR #373, opened 2024-04-20
Update to go-client 0.26.15

pjonsson:
This removes some dependencies
on older golang.org/x/net which
contain CVE-2022-41723 and
CVE-2023-39325.


pjonsson:
@rjbez17 it's been 2 months since I updated the PR to include `go mod vendor` results, could you please take another look?

PR #374, opened 2024-04-20
Update to controller-tools 0.14.0

pjonsson:
This updates golang.org/x/tools to
a later version which fixes the crash
in https://github.com/kubernetes-sigs/controller-tools/issues/880, which I believe is what is happening
on the current main branch in this
repository.

I'm really after the security fixes, so pick #373 if that smaller change is preferable to this.


pjonsson:
The build failure in the test is because controller-runtime is < v0.15.0 and some version of something in a dependency isn't updated yet. Controller-tools 0.15 should not have this problem, but my reading of the Compatibility section at https://pkg.go.dev/sigs.k8s.io/controller-tools#section-readme is that it requires Go 1.22.

@rjbez17 I know this project uses Go 1.20, do you have any advice?

pjonsson:
@rjbez17 I don't really know Go, but let me get back to you on that next week. And Go 1.21 doesn't help us for this particular issue, but I'm guessing starting with that will be easier since it is a smaller change.

Can I get the CI to trigger without your explicit approval, or do I need to join the kubernetes-sigs to make that happen?

rjbez17:
The ok-to-test label should trigger the tests now. You can use /test all in a comment as well

rjbez17:
And thanks for taking a look!

pjonsson:
@rjbez17 Considering I have no idea what I'm doing, I started with Go 1.21 in #406.

I will take a look at 1.22 tomorrow, but that seems more difficult, so if you believe the 1.21 PR works I suggest start with merging that.

Even if Go 1.21 and the latest main branch "just" fixes a few of the security issues that Trivy is complaining about, we're closing up on 2 years since the last release so it would be nice with a release sometime soonish too.

rjbez17:
> @rjbez17 Considering I have no idea what I'm doing, I started with Go 1.21 in #406.
> 
> 
> 
> I will take a look at 1.22 tomorrow, but that seems more difficult, so if you believe the 1.21 PR works I suggest start with merging that.
> 
> 
> 
> Even if Go 1.21 and the latest main branch "just" fixes a few of the security issues that Trivy is complaining about, we're closing up on 2 years since the last release so it would be nice with a release sometime soonish too.

I agree and really appreciate your help here. Once we get this merged I'll try to find some time to time box a 1.22 upgrade and if not release a maintenance release within the next week or so.

PR #377, opened 2024-05-10
feat: Parametrize number of replicas for HA deployment

mjozefcz:
For some deployments 3 replicas are a bit too much. Let's parametrize it.


rjbez17:

PR #380, opened 2024-05-23
feat(chart): create & mount service account

Nold360:
this PR adds the ability to use / create a serviceAccount != "default"

### DIFF
--- a/charts/hnc/Chart.yaml
+++ b/charts/hnc/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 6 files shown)


mochizuki875:
@Nold360 

Thank you for your suggestion and PR!

There is just one thing to note.
Now, to generate Helm Chart, we need to execute [`make charts`](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#install-with-helm).
It is for keeping up-to-date with changes in HCN manifests(`manifests/*.yaml` generated by `make manifests`), which was  barrier to create Helm chart [previously](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/46#issuecomment-1013224494).

Then, all files in `charts/hnc/templates` are re-generated by `make charts` [like this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/a0351d29a158c0a0923db69ee6010c65cdf59ad2/Makefile#L512-L517), so you should not edit `charts/hnc/templates/xxx.yaml` directly.
Sorry, I should left comment somewhere...

So in this case, it's good to update [`hack/helm_patches/update-helm.sh`](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/hack/helm_patches/update-helm.sh) to update template.

Issue #312, opened 2023-07-03
excluding all default namespaces 


zfrhv:
hello~
we have openshift cluster so i was configuring the HNC to exclude default openshift namespaces
but its like 70 namespaces (plus each redhat operator will create additional namesoace)

so we configured `--included-namespace-regex=(\?\!openshift).*`, but looking at how kubernetes is excluded im wondering, why to exclude each seperate namespace and not just do wildcard same as i did?:
instead if [this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/config/manager/manager.yaml#L48-L51):
```yaml
        - "--excluded-namespace=kube-system"
        - "--excluded-namespace=kube-public"
        - "--excluded-namespace=hnc-system"
        - "--excluded-namespace=kube-node-lease"
```
do this:
```yaml
        - "--included-namespace-regex=(\?\!kube-|openshift).*"
        - "--excluded-namespace=kube-system"
```

also without the backslashes it was crashing, not sure if its a feature or a bug

PR #382, opened 2024-06-10
exclude openshift namespaces via regex

zfrhv:
Adding `--included-namespace-regex` option to the helm chart.  
as mentioned [in the docs](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#including-and-excluding-namespaces-from-hnc) there is only `--included-namespace-regex` or `--excluded-namespace`.  
so in order to exclude all `openshift.*` namespaces I used the `--included-namespace-regex` as mentioned [in this issue](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/312), this is added as an example in `values.yaml`

### DIFF
--- a/charts/hnc/templates/hnc-admin.yaml
+++ b/charts/hnc/templates/hnc-admin.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+  name: '{{ include "hnc.fullname" . }}-admin'
+rules:
+  - apiGroups:
+      - hnc.x-k8s.io
+    resources:
+      - hierarchicalresourcequotas
+      - subnamespaceanchors
+      - hierarchyconfigurations
+    verbs:
+      - '*'
--- a/charts/hnc/templates/hnc-controller-manager-ha.yaml
+++ b/charts/hnc/templates/hnc-controller-manager-ha.yaml
@@ -23,7 +23,10 @@ spec:
             {{- if .Values.hrq.enabled }}
             - --enable-hrq
             {{- end }}
-            {{- range $hncExcludeNamespace := .Values.hncExcludeNamespaces}}
+            {{- if $hncIncludeNamespacesRegex }}
+            - --included-namespace-regex={{ $hncIncludeNamespacesRegex }}
+            {{- end }}
+            {{- range $hncExcludeNamespace := .Values.hncExcludeNamespaces }}
             - --excluded-namespace={{ $hncExcludeNamespace }}
             {{- end }}
             - --webhook-server-port=9443
--- a/charts/hnc/templates/hnc-controller-manager.yaml
+++ b/charts/hnc/templates/hnc-controller-manager.yaml
@@ -22,7 +22,10 @@ spec:
             {{- if .Values.hrq.enabled }}
             - --enable-hrq
             {{- end }}
-            {{- range $hncExcludeNamespace := .Values.hncExcludeNamespaces}}
+            {{- if $hncIncludeNamespacesRegex }}
+            - --included-namespace-regex={{ $hncIncludeNamespacesRegex }}
+            {{- end }}
+            {{- range $hncExcludeNamespace := .Values.hncExcludeNamespaces }}
             - --excluded-namespace={{ $hncExcludeNamespace }}
             {{- end }}
             {{- if .Values.ha.enabled }}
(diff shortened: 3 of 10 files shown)


zfrhv:
@mochizuki875 can you please tell if you like it?

mochizuki875:
@zfrhv 

Thank you for your suggestion and PR!

There is just one thing to note.
Now, to generate Helm Chart, we need to execute [`make charts`](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#install-with-helm).
It is for keeping up-to-date with changes in HCN manifests(`manifests/*.yaml` generated by `make manifests`), which was  barrier to create Helm chart [previously](https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/46#issuecomment-1013224494).

Then, all files in `charts/hnc/templates` are re-generated by `make charts` [like this](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/Makefile#L512-L517), so you should not edit `charts/hnc/templates/xxx.yaml` directly.
Sorry, I should left comment somewhere...

So in this case, it's good to update `hack/helm_patches/update-helm.sh` to update template like that.
https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/hack/helm_patches/update-helm.sh#L176-L177

zfrhv:
@mochizuki875 
thank you for mentioning.
I updated the files accordingly, after running `make charts` it did updated the templates as expected.
i didnt added the `make charts` result to the PR because it also modified many other files

let me know if you want me to include the `make charts` result in the PR

mochizuki875:
@zfrhv 
IMO, I think it's better to add `make charts` result if it works as expected, because I think latest charts are expected to be in `charts` dir of project and some users may use the chart without `make charts`. 

Ideally, `make charts` should be executed automatically and the helm chart should be updated to `charts` dir and github pages as commented [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/354#issuecomment-2009007851), but I’ve not started working yet...

zfrhv:
in values.yaml the image repository seems to be broken
am I supposed to dig in and figure out why?

mochizuki875:
Have you set some environment values? ([see here](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#install-with-helm))

By setting these values as below,

```bash
$ export HNC_REGISTRY=gcr.io/k8s-staging-multitenancy
$ export HNC_IMG_NAME=hnc-manager
$ export HNC_IMG_TAG=v1.1.0 
```

values.yaml should be generated like this as a result of `make charts`.
```yaml
image:
  repository: gcr.io/k8s-staging-multitenancy/hnc-manager
  tag: v1.1.0
  imagePullPolicy: {}
  ...
```

zfrhv:
sorry for making you feed me with a spoon

mochizuki875:
@zfrhv 
Thank you for your work!
This PR seems to make sense.
/lgtm
/cc @rjbez17

Related this PR, I have two questions.
If you have some information, please let me know.

1. Some ClusterRoles(admin, edit, view) templates are replaced from admin-role in this PR.
However, there doesn’t seems to be Binding resource to bind them.
Is this intentional?
It comes from #318.

2. (NIT) At first, I thought the range operation should be used for `--included-namespace-regex` like [--excluded-namespace](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/977e2c533a0efafddb0d2e40eeb8e69e9f3a0c30/charts/hnc/templates/hnc-controller-manager.yaml#L25-L27).
However, `includedNamespacesRegex` which comes from `--included-namespace-regex` is [defined as string](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/977e2c533a0efafddb0d2e40eeb8e69e9f3a0c30/cmd/manager/main.go#L78) not arrayArg, so we can't.
How do you think?

zfrhv:
@mochizuki875 
1. those roles are the default permissions that users will get.
  for example if someone has view on his namespace, then he will also be able to view the hnc resources on his namespace.
  if someone has edit then he can edit and also admin.
  the clusterRoles are not supposed to be bound to anything, they have the label `rbac.authorization.k8s.io/aggregate-to-edit: "true"` which kubernetes uses to merge the cluster role into the default `clusterRole edit`.
  its just better permissions assignment instead of just giving admin to admin
2. im not good at go but when U checked how to use the includeNamespaceRegex I saw the pr that added it, and it mentions that its a string and not arrayArg [here](https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/89/files) lines 69-70

mochizuki875:
@zfrhv Thanks your information;)

Issue #385, opened 2024-08-12
helm extra controller args


zfrhv:
hi @mochizuki875
I want to add --unpropagated-label=<string> to the helm chart as mentioned in the docs:
https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#admin-managed-labels

should I add something like:
```yaml
manager:
  extraArgs:
  - --unpropagated-label: app.kubernetes.io/instance
```
or something like:
```yaml
manager:
  unpropagatedLabel: app.kubernetes.io/instance
```
?

because there are many other controller arguments

mochizuki875:
@zfrhv 
Sorry for late a response...
IMO, it's preferable like that.

```yaml
manager:
  unpropagatedLabel:
  - app.kubernetes.io/instance
  - example.com/foo
```

Additionally, it might be beneficial to handle `--unpropagated-annotation` similarly, as it appears to have similar characteristics to `--unpropagated-label`.

PR #386, opened 2024-08-23
feat: add chart values to manage labels and annotations

kalexmills:
Without the ability to pass additional command-line arguments to the manager, installing HNC using a tool like ArgoCD requires developers to maintain their own fork.

This is a targetted PR relating to #385 for the special case of managed annotations and labels.

A separate PR could be used in lieu of this change, for passing any and all 'extraArgs' to the HNC controller that may be needed, now or in future. #387. I'll leave the choice of which to merge to the maintainers.


rjbez17:

PR #387, opened 2024-08-23
feat: extra args on manager

kalexmills:
Without the ability to pass additional command-line arguments to the manager, installing HNC using a tool like ArgoCD requires developers to maintain their own fork.

This PR is an alternative implementation that's closer to the spirit of #385, and mirrors pattterns used in other helm charts for passing extra arguments to commands.

Related to #386. Both could be merged, I guess, but this one is all that's absolutely necessary.


rjbez17:

PR #406, opened 2025-03-01
Update to Go 1.21

pjonsson:
Bump the Go versions in various
files, and run go mod tidy to
clean up things.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM golang:1.21 AS builder
 
 WORKDIR /workspace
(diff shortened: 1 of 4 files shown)


pjonsson:
This passes `go test ./api/... ./cmd/... ./internal/... -coverprofile cover.out` in a Go 1.21 image for me locally, and the Dockerfile builds. I hijacked #374 where I already had ok-to-test and this passed the pull-hnc-test job there. Not sure how the pull-hnc-test job could pass since that uses a Go 1.20 image, but it did.

Filed https://github.com/kubernetes/test-infra/pull/34440 to fix presubmit.

PR #408, opened 2025-03-02
Update to Go 1.22

pjonsson:
Bump the Go versions in various
files and run go mod tidy to clean
up things, Then update to
controller-tools 0.15 to fix
the segmentation fault when running
controller-gen.

Being able to run controller-gen
gives a number of new compilation
errors, so update the K8S packages
to 0.27.16, controller-runtime to
0.15.3 which supports Kubernetes 1.27,
and open-policy-agent/cert-controller
to 0.8.0 which also supports
Kubernetes 1.27.


internal/anchor/reconciler.go
```diff
@@ -424,7 +424,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&api.SubnamespaceAnchor{}).
-		Watches(&source.Channel{Source: r.affected}, &handler.EnqueueRequestForObject{}).
-		Watches(&source.Kind{Type: &corev1.Namespace{}}, handler.EnqueueRequestsFromMapFunc(nsMapFn)).
+		WatchesRawSource(&source.Channel{Source: r.affected}, &handler.EnqueueRequestForObject{}).
```
pjonsson, line 427: I know one is supposed to use Watches(), but I don't understand the builder stuff and how to update these callers for the new type-safe API.

internal/config/client.go
```diff
@@ -2,37 +2,40 @@ package config
...
-	})
+	return c, nil
 }
 
-func NewClient(readOnly bool) cluster.NewClientFunc {
-	return func(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
-		c, err := newCachingClient(cache, config, options, uncachedObjects...)
+func NewClient(readOnly bool) client.NewClientFunc {
+	return func(config *rest.Config, options client.Options) (client.Client, error) {
+		// FIXME: mgr.GetCache() gives us the cache, but this is called when creating
```
pjonsson, line 35: Need help with this one. To me it looks like we need the cache that we get from the manager, while we are in the process of creating the manager.

I'm fairly sure the idea conveyed with my code changes in `newCachingClient()` is correct though, it's now possible to create a caching client by passing the right options to `client.New()`.

internal/config/client.go
```diff
@@ -94,6 +97,16 @@ func (c readOnlyClient) RESTMapper() meta.RESTMapper {
 	return c.client.RESTMapper()
 }
 
+func (c readOnlyClient) GroupVersionKindFor(obj runtime.Object) (schema.GroupVersionKind, error) {
+	// FIXME: Broken implementation.
```
pjonsson, line 101: Need help with these two implementations as well, and there's a second set of broken implementations of these functions in `validator_test.go`.

internal/hncconfig/validator.go
```diff
@@ -227,7 +228,9 @@ func (a ancestorObjects) add(onm string, ns *forest.Namespace) {
 }
 
 func (v *Validator) InjectConfig(cf *rest.Config) error {
-	mapper, err := apimeta.NewGroupKindMapper(cf)
+	// FIXME: Where do we get the client from?
+	client := &http.Client{Timeout: 180}
```
pjonsson, line 232: I'm guessing we can't change the signature of injection functions, so I have no idea how to get the `http.Client` at this point.

pjonsson:
@rjbez17 I tried to make the minimal change for getting Go 1.22 going (calling +82k-61k lines minimal doesn't feel right). This builds in a Go 1.22 container locally for me, but there are things I don't know how to change so I've made comments on them in this PR that I was hoping you could help me out with.

To get the post-submit tests to not crash require a more recent controller-tools (or controller-runtime, I don't remember), but that brings in a slew of other dependencies that also need to be updated along with some API changes that requires changes in the HNC code. I have some of those updates in a local branch (that doesn't build), but I'm trying to not pile on everything at once.

The bot told me the other PR I was using for testing in the CI needed to be rebased; not sure how I broke that because it worked yesterday, but it might be related to the Go 1.21 update being merged to the main branch. Could you please give this PR the ok-to-test so I could use this PR for the rest?

I also prepared the test-infra bump for later: https://github.com/kubernetes/test-infra/pull/34442

internal/config/client.go
rjbez17: The release notes [here](https://newreleases.io/project/github/kubernetes-sigs/controller-runtime/release/v0.15.0)

> Add client.{GroupVersionKindFor, IsObjectNamespaced} ([#2136](https://github.com/kubernetes-sigs/controller-runtime/pull/2136))
> 
>     The client.Client interface now requires and exposes these helper functions:
>         GroupVersionKindFor(Object) which returns the schema.GroupVersionKind for the object.
>         IsObjectNamespaced(Object) which returns true if the object's GroupVersionKind is namespaced, or false for global ones.

So I think we can follow suit of the other func and use:

```
func (c readOnlyClient) GroupVersionKindFor(obj runtime.Object) (schema.GroupVersionKind, error) {
	return c.client.GroupVersionKindFor(obj)
}
```

pjonsson: Oh, that was easy, fixed! Thank you!

internal/objects/validator_test.go
```diff
@@ -721,6 +721,12 @@ func (fakeNSClient) RESTMapper() meta.RESTMapper {
 func (fakeNSClient) Scheme() *runtime.Scheme {
 	return nil
 }
+func (fakeNSClient) GroupVersionKindFor(_ runtime.Object) (schema.GroupVersionKind, error) {
+	return schema.GroupVersionKind{}, nil
+}
+func (fakeNSClient) IsObjectNamespaced(_ runtime.Object) (bool, error) {
+	return true, nil
```
pjonsson, line 728: No client in `fakeNSClient`, so these still return dummy values. Not sure if `true` is a good default for this function.

internal/hncconfig/validator.go
pjonsson: The previous comment doesn't matter, because the inject functions are no longer used according to the release notes (https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0).

My impression is that the inject stuff should be moved to the construction of the validator, but this turned into a skill issue. Where are the validators constructed?

internal/config/client.go
pjonsson: Ok, I removed the constant `nil` parameter and the FIXME. The `uncachedObjects` argument will never be passed, but I guess that is ok.

internal/objects/validator_test.go
rjbez17: Seems like an ok default to me

internal/anchor/reconciler.go
rjbez17: `WatchesRawSource` seems ok (for now at least). Refering again to the release notes its the prefered way to use `source` and since we aren't using predicates I think we should be fine here:

> Improve builder package capabilities and general UX ([#2135](https://github.com/kubernetes-sigs/controller-runtime/pull/2135))
> 
>     builder.Builder.Watches signature has changed, it now takes a client.Object instead of a source.Source as first argument.
>         For source.Source, use WatchesRawSource.

rjbez17:
Thank you so much for this PR and all the work you've put in. I think I've covered all of your questions so far. I will set aside some time either later tonight or no later than tomorrow night to do a deeper dive and test this out locally. Once we get this merged in and tested I'll cut a maintenance release.

pjonsson:
@rjbez17 sorry, I wasn't very clear, but https://github.com/kubernetes-sigs/hierarchical-namespaces/pull/408#discussion_r1977634355 still confuses me. I realize the (long deprecated) injection API is gone so I removed the injection functions, but I don't know how to restore the functionality that removed.

I'm guessing the missing injections are the cause for the current simple test failures in my local container. I think the crash on main for the post-submit-job (E2E-tests?) require the controller-tools update that #374 was meant to solve before I ran into the Go 1.22 requirement.

I suspect fixing this stuff and everything else is a 30 minute thing for someone who knows Go, Kubernetes, and this code base, so you're more than welcome to take this across the finish line.

If you decide to update dependencies further than this PR, the "big" thing I had in my non-building branch is changing `*admission.Decoder` into `admission.Decoder`, because that was changed to an interface (or something like that) in some later release of one of the dependencies.

pjonsson:
@rjbez17 I realize you're busy so I didn't want to disturb you the same week, but it's been a couple of weeks now, did other things get in the way of your plans?

rjbez17:
Hey sorry about this. I tried reaching out in k8s slack. We had a family emergency and I've been out last week. I'll take a look before end of week. Again, so sorry to keep you hanging, I appreciate all the help.

pjonsson:
@rjbez17 I'm sorry to hear that. A few more weeks here or there won't make a big difference, take care of family first and then return to this when the important stuff is taken care of.

